PR #5, opened 2021-02-22
Implemented My Kentik API get and get all methods

lukaa12:
KNTK-177


examples/my_kentik_portal_example_test.go
```diff
@@ -0,0 +1,49 @@
...
+	runCRUDexample()
+	getAllTenants()
+}
+
+func runCRUDexample() {
+	client := NewClient()
+
+	fmt.Println("### GET")
+	tenant, err := client.MyKentikPortal.Get(context.Background(), 577)
+	PanicOnError(err)
```
mateuszmidor, line 26: today's morning with Daniel we agreed on not panicking but returning errors which is more verbose but also more GOthic :)

kentikapi/internal/api_payloads/tenant.go
```diff
@@ -0,0 +1,79 @@
...
+		return models.Tenant{}, err
+	}
+	var companyID models.ID
+	if len(users) != 0 {
+		companyID = p.Users[0].CompanyID
+	}
+	return models.Tenant{
+		ID:          p.ID,
+		CompanyID:   companyID,
+		Name:        &p.Name,
```
mateuszmidor, line 53: not sure: why is Tenant.Name pointer to string, but TenantPayload.Name just string? It is optional field or not?

lukaa12: It should be string. It's not optional.

mateuszmidor (dismissed):

kentikapi/internal/api_resources/my_kentik_portal_test.go
```diff
@@ -0,0 +1,228 @@
...
+	tenants, err := myKentikPortalAPI.GetAll(nil)
+
+	//assert
+	assert := assert.New(t)
+	require := require.New(t)
+
+	require.NoError(err)
+	assert.Zero(transport.RequestBody)
+
+	assert.Len(tenants, 2)
```
Fenthick, line 65: I think that it would be more readable to check response with single assert statement: `assert.Equal(expectedTenant, tenant)` 

In other tests too.

See: https://github.com/kentik/community_sdk_golang/blob/7ed66588c532214584238df5a8dba31e0c7b0e9b/kentikapi/client_integration_test.go#L170

kentikapi/internal/api_resources/my_kentik_portal_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	//act
+	tenantUser, err := myKentikPortalAPI.CreateTenantUser(nil, 577, "test@test.test")
+
+	//assert
+	assert := assert.New(t)
+	require := require.New(t)
+
+	require.NoError(err)
+	payload := utils.NewJSONPayloadInspector(t, transport.RequestBody)
```
Fenthick, line 200: I think that we should check that API does not send redundant data, e.g.
```json
{
  "user": {
    "user_email": "foo",
    "redundant_trash": "trash",
  }
}
```

I suggest to check number of keys in JSON objects.

kentikapi/internal/api_resources/my_kentik_portal_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+	//act
+	tenants, err := myKentikPortalAPI.GetAll(nil)
+
+	//assert
+	assert := assert.New(t)
+	require := require.New(t)
+
+	require.NoError(err)
+	assert.Zero(transport.RequestBody)
```
Fenthick, line 63: Please validate the request path passed to transport.
In other tests too.

kentikapi/internal/api_resources/my_kentik_portal_test.go
```diff
@@ -0,0 +1,228 @@
...
+	transport := &api_connection.StubTransport{ResponseBody: deleteResponsePayload}
+	myKentikPortalAPI := api_resources.NewMyKentikPortalAPI(transport)
+
+	// act
+	tenantID := models.ID(478)
+	userID := models.ID(420)
+	err := myKentikPortalAPI.DeleteTenantUser(nil, tenantID, userID)
+
+	// assert
+	assert := assert.New(t)
```
Fenthick, line 224: There is no need to assertion objects in such tests.
You could just use `assert.Zero(t, foo)` forms, etc.

Fenthick (changes requested):
Good job overall! I suggested several improvements.

kentikapi/internal/api_resources/my_kentik_portal_test.go
mateuszmidor: API does return more fields than we deserialize. They are described as "internal-use" in the doc and we consciously ignore them.

Fenthick: I meant the request payload that the library sends (API client, not API).

kentikapi/internal/api_payloads/tenant.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (p TenantPayload) ToTenant() (models.Tenant, error) {
+	var users []models.TenantUser
+	err := utils.ConvertList(p.Users, TenantUserPayload.ToTenantUser, &users)
+	if err != nil {
+		return models.Tenant{}, err
+	}
+	var companyID models.ID
+	if len(users) != 0 {
+		companyID = p.Users[0].CompanyID
```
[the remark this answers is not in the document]
lukaa12: It is here: https://github.com/kentik/community_sdk_python/blob/bbe61cc9c8257b945ac65825ea1eb7531bf0763f/kentik_api_library/kentik_api/public/tenant.py#L37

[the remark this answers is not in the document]
Fenthick: Ok, I see that the documentation (https://kb.kentik.com/v0/Ec09.htm#Ec09-Tenant_Object) is inconsistent regarding the tenant.company_id and the API does not return that field. Let's keep that logic.
Resolved

kentikapi/internal/api_resources/my_kentik_portal_test.go
lukaa12: Added TODO

kentikapi/internal/api_payloads/tenant.go
```diff
@@ -0,0 +1,78 @@
...
+	Description string              `json:"description"`
+	Users       []TenantUserPayload `json:"users"`
+	CreatedDate time.Time           `json:"created_date"`
+	UpdatedDate time.Time           `json:"updated_date"`
+}
+
+type TenantUserPayload struct {
+	ID        models.ID  `json:"id,string"`
+	Email     string     `json:"user_email"`
+	Name      *string    `json:"user_name"`
```
mateuszmidor, line 32: add omitempty for pointers

mateuszmidor (approved):

kentikapi/internal/utils/convert.go
```diff
@@ -80,3 +86,11 @@ func ConvertOrNone(input interface{}, convertFunc ConvertFunc, output interface{
 	vOutput.Elem().Set(results[0])
 	return nil
 }
+
+func ParseISO8601Timestamp(t testing.TB, timestamp string) *time.Time {
```
Fenthick, line 90: I think we should separate testing code from implementation.
Please put this function to separate testutil package just like here: https://github.com/kentik/community_sdk_golang/blob/6d0df03855e249bd2a28edf790a6bab17eb63f01/kentikapi/internal/testutil/format.go

kentikapi/internal/api_resources/my_kentik_portal_test.go
Fenthick: @lukaa12 Please add a TODO for that, e.g.
`TODO: Verify that that there is no redundant data sent in request body`

It is important to check that we do not send harmful data that would e.g. zero value of a field - set it to `""` or `0`.

kentikapi/internal/api_resources/my_kentik_portal_test.go
Fenthick: Resolved

kentikapi/internal/utils/convert.go
lukaa12: I reconsidered using this function in my_kentik_portal api tests, and leave it untouched compared to master branch.

kentikapi/internal/api_resources/my_kentik_portal_test.go
lukaa12: TODO added

kentikapi/internal/utils/convert.go
Fenthick: Great. Resolved

Fenthick (approved):

PR #9, opened 2021-02-24
Run tests and linters with Github Actions CI

Fenthick:
Use Super-Linter to run multiple static analysis tools for multiple files.
Use golangci-lint for the Go code
Fix code to comply to linters

Note that "test" job could use Actions caching functionality to cache downloaded
Go modules and Go build cache. This improvement could be configured in
the future.

Issue: KNTK-182

### DIFF
--- a/kentikapi/internal/api_payloads/device.go
+++ b/kentikapi/internal/api_payloads/device.go
@@ -336,19 +336,6 @@ func payloadToDeviceLabel(p deviceLabelPayload) (models.DeviceLabel, error) {
 	}, nil
 }
 
-// nolint: deadcode, unused
-func deviceLabelToPayload(d models.DeviceLabel) (deviceLabelPayload, error) {
-	return deviceLabelPayload{
-		ID:          d.ID,
-		Name:        d.Name,
-		Color:       d.Color,
-		UserID:      d.UserID,
-		CompanyID:   d.CompanyID,
-		CreatedDate: d.CreatedDate,
-		UpdatedDate: d.UpdatedDate,
-	}, nil
-}
-
 // deviceSitePayload represents JSON Device.Site payload as it is transmitted from KentikAPI.
 // deviceSitePayload embeddedd under Device differs from regular SitePayload in that all fields are optional.
 type deviceSitePayload struct {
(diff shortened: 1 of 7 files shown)


Fenthick:
Here you can see the output of the workflow: https://github.com/Fenthick/community_sdk_golang/actions/runs/599250103

kentikapi/internal/api_payloads/device.go
```diff
@@ -336,6 +336,7 @@ func payloadToDeviceLabel(p deviceLabelPayload) (models.DeviceLabel, error) {
 	}, nil
 }
 
+// nolint: deadcode, unused
```
Fenthick, line 339: @mateuszmidor Is it going to be used in the near future? Otherwise I'll remove the function

mateuszmidor: You can remove it.

kentikapi/internal/api_resources/devices_test.go
```diff
@@ -112,7 +113,7 @@ func TestCreateDeviceRouter(t *testing.T) {
 	models.SetOptional(&router.DeviceBGPPassword, "bgp-optional-password")
 	models.SetOptional(&router.SiteID, 8483)
 	models.SetOptional(&router.DeviceBGPFlowSpec, true)
-	device, err := devicesAPI.Create(nil, *router)
```
mateuszmidor, line 115: StubTransport doesnt use the context

kentikapi/internal/api_payloads/device.go
Fenthick: Done

mateuszmidor (approved):

kentikapi/internal/api_resources/devices_test.go
Fenthick: Ok, but I did this change to suppress the linter output. And I guess this check is based on the Context package documentation (https://pkg.go.dev/context@go1.16#pkg-overview):
> Do not pass a nil Context, even if a function permits it. Pass context.TODO if you are unsure about which Context to use. 

Do you suggest to disable this check globally / locally?
In my opinion the best option here is just to pass `Background()` here and move on.

mateuszmidor: agree, no point in fighting the tool here

lukaa12 (approved):

TedatKentik (approved):
updating tests and linters

PR #16, opened 2021-03-04
Implemented Saved Filter API

lukaa12:
KNTK-175


examples/saved_filters_example_test.go
```diff
@@ -0,0 +1,92 @@
...
+	fmt.Println()
+
+	return nil
+}
+
+func runCRUDSavedFilters() error {
+	client := NewClient()
+
+	fmt.Println("### GET")
+	savedFilter, err := client.SavedFilters.Get(context.Background(), 8275)
```
mateuszmidor, line 38: May be better to call GET after CREATE, with the created filter ID. So the ID always exists.

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,182 @@
...
+	SavedFilterPayload
+}
+
+type SavedFilterPayload struct {
+	ID                IntAsString    `json:"id,omitempty"`
+	CompanyID         models.ID      `json:"company_id,string,omitempty"`
+	FilterName        string         `json:"filter_name"`
+	FilterDescription string         `json:"filter_description"`
+	FilterLevel       string         `json:"filter_level,omitempty"`
+	CreatedDate       *time.Time     `json:"cdate,omitempty"`
```
mateuszmidor, line 47: You can validate that payload received from API contains CreatedDate,UpdatedDate fields with struct tags response:"get,post,put" if they are expected to always be present. So dereferencing these fields with * doesnt end up with seg fault in case null is received

kentikapi/internal/api_resources/saved_filters.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return response.ToSavedFilters()
+}
+
+func (a *SavedFiltersAPI) Get(ctx context.Context, filterID models.ID) (*models.SavedFilter, error) {
+	var response api_payloads.GetSavedFilterResponse
+	err := a.GetAndValidate(ctx, api_endpoints.GetSavedFilter(filterID), &response)
+	if err != nil {
+		return &models.SavedFilter{}, err
```
mateuszmidor, line 33: You can just: return nil, err

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,182 @@
...
+type UpdateSavedFilterRequest struct {
+	SavedFilterPayload
+}
+
+type UpdateSavedFilterResponse struct {
+	SavedFilterPayload
+}
+
+type SavedFilterPayload struct {
+	ID                IntAsString    `json:"id,omitempty" response:"get,post,put"`
```
Fenthick, line 42: Response validation is performed on deserialized response and checks if the field is nil pointer/interface. If there is no e.g. filter_name in response, the field will have `""` value the validation will not detect that. This could be misleading. Because of that I propose to keep response validation only for pointer/interface types.

So for this payload it makes sense to have those tags only for CreatedDate and UpdatedDate.

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,182 @@
...
+		FilterName:        sf.FilterName,
+		FilterDescription: sf.FilterDescription,
+		FilterLevel:       sf.FilterLevel,
+		CreatedDate:       &sf.CreatedDate,
+		UpdatedDate:       &sf.UpdatedDate,
+		Filters:           FiltersToPayload(sf.Filters),
+	}
+}
+
+type FiltersPayload struct {
```
Fenthick, line 83: I think that in this payload no request/response validation tags make sense. See explanation above.

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,182 @@
...
+
+	return FiltersPayload{
+		Connector:    f.Connector,
+		Custom:       f.Custom,
+		FilterGroups: filterGroups,
+		FilterString: f.FilterString,
+	}
+}
+
+type FilterGroupsPayload struct {
```
Fenthick, line 118: I think that in this payload no request/response validation tags make sense. See explanation above.

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,182 @@
...
+
+type UpdateSavedFilterRequest struct {
+	SavedFilterPayload
+}
+
+type UpdateSavedFilterResponse struct {
+	SavedFilterPayload
+}
+
+type SavedFilterPayload struct {
```
Fenthick, line 41: Nit: I think that payload structs, such as SavedFilterPayload could (and should) be unexported in order to reduce package API surface.

kentikapi/internal/api_resources/saved_filters_test.go
```diff
@@ -0,0 +1,3 @@
+package api_resources_test
+
+// TODO(lwolanin) Implement tests for Saved Filters API
```
Fenthick, line 3: Please do that as part of this PR.

Fenthick (changes requested):

kentikapi/models/saved_filter.go
```diff
@@ -0,0 +1,37 @@
+package models
+
+import "time"
+
+type SavedFilter struct {
```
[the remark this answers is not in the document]
lukaa12: Done

kentikapi/internal/api_resources/saved_filters_test.go
lukaa12: Done

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -79,16 +80,46 @@ func SavedFilterToPayload(sf models.SavedFilter) SavedFilterPayload {
 	}
 }
 
-type FiltersPayload struct {
+func SavedFilterToCreatePayload(sf models.SavedFilter) CreateSavedFilterRequest {
+	return CreateSavedFilterRequest{
+		savedFilterPayload: savedFilterPayload{
```
Fenthick, line 85: Please use SavedFilterToPayload() here to remove duplication.
Alternatively remove  SavedFilterToCreatePayload() and SavedFilterToUpdatePayload() function and use only SavedFilterToPayload() everywhere.

For SavedFilterToUpdatePayload function too.

kentikapi/internal/api_payloads/query.go
```diff
@@ -240,20 +240,26 @@ func queryToPayload(q models.Query) (queryPayload, error) {
 }
 
 type filtersPayload struct {
-	Connector    string        `json:"connector"`
-	FilterGroups []interface{} `json:"filterGroups,omitempty"` // []FilterGroupsPayload; To be implemented in SavedFiltersAPI
-	Custom       *bool         `json:"custom,omitempty"`
-	FilterString *string       `json:"filterString,omitempty"`
+	Connector    string                `json:"connector"`
+	FilterGroups []FilterGroupsPayload `json:"filterGroups,omitempty"` // []FilterGroupsPayload; To be implemented in SavedFiltersAPI
```
mateuszmidor: it's nice you spotted this :+1:

kentikapi/internal/api_resources/saved_filters_test.go
```diff
@@ -1 +1,360 @@
...
+				},
+			},
+		},
+		CompanyID:   74333,
+		FilterLevel: "company",
+		CreatedDate: time.Date(2020, 12, 16, 10, 46, 13, 95e6, time.UTC),
+		UpdatedDate: time.Date(2020, 12, 26, 10, 46, 13, 95e6, time.UTC),
+		ID:          8152,
+	}
+	expectedRequestPayload := "{\"filter_name\":\"test_filter1\",\"filter_description\":\"This is test filter description\",\"cdate\":\"0001-01-01T00:00:00Z\"," +
```
Fenthick, line 252: This is not readable for me. And it also tests the order of keys in the JSON, which is not wanted and too strict. I think the better approach is to use:
1. JSONPayloadInspector like in most of tests (we have a ticket to make it more strict, BTW: https://clkentik.atlassian.net/browse/KNTK-183)
2. unmarshal transport.RequestBody to interface{} and compare with expected map[string]interface{} (see: tags_integration_test.go -> TestClient_CreateTag)

The same in TestUpdateSavedFilter 

You could add TODO if you feel you're time-restricted.

kentikapi/internal/api_payloads/query.go
```diff
@@ -209,11 +209,16 @@ func queryToPayload(q models.Query) (queryPayload, error) {
...
+	for _, i := range q.SavedFilters {
+		savedFiltersPayloads = append(savedFiltersPayloads, SavedFilterToPayload(i))
+	}
+
 	return queryPayload{
 		Metric:          q.Metric.String(),
 		Dimension:       dimensions,
 		FiltersObj:      filtersToPayload(q.FiltersObj),
-		SavedFilters:    []interface{}{}, // SavedFiltersAPI dependency
+		SavedFilters:    savedFiltersPayloads, // SavedFiltersAPI dependency
```
mateuszmidor, line 221: comments regarding SavedFiltersAPI dependency can be removed

examples/saved_filters_example_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"testing"
+
+	"github.com/kentik/community_sdk_golang/kentikapi/models"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSavedFiltersAPIExample(t *testing.T) {
+	assert := assert.New(t)
+	assert.NoError(runCRUDSavedFilters())
+	// assert.NoError(runGetAllSavedFilters())
```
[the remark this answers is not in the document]
Fenthick: Done

kentikapi/models/saved_filter.go
```diff
@@ -3,35 +3,56 @@ package models
...
-	UpdatedDate       time.Time
 	Filters           Filters
+
+	// read-only properties (can't be updated in update call)
+	ID          ID
+	CompanyID   ID
+	CreatedDate time.Time
+	UpdatedDate time.Time
+}
+
```
Fenthick, line 18:
```suggestion

// NewSavedFilter creates a new SavedFilter with all required fields set.
```

kentikapi/models/saved_filter.go
```diff
@@ -0,0 +1,37 @@
...
+	ID                ID
+	CompanyID         ID
+	FilterName        string
+	FilterDescription string
+	FilterLevel       string
+	CreatedDate       time.Time
+	UpdatedDate       time.Time
+	Filters           Filters
+}
+
```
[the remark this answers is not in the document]
Fenthick: Done

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: I see that ID and Company ID still have validation tags:
```
ID                IntAsString    `json:"id,omitempty" response:"get,post,put"`
CompanyID         models.ID      `json:"company_id,string,omitempty" response:"get,post,put"`
```

Aren't they redundant?

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -79,16 +80,46 @@ func SavedFilterToPayload(sf models.SavedFilter) SavedFilterPayload {
...
+			Filters:           FiltersToPayload(sf.Filters),
+		},
+	}
+}
+
+type filtersPayload struct {
 	Connector    string                `json:"connector"`
 	Custom       *bool                 `json:"custom,omitempty"`
-	FilterGroups []FilterGroupsPayload `json:"filterGroups"`
+	FilterGroups []filterGroupsPayload `json:"filterGroups" response:"get,post,put"`
```
Fenthick, line 116: Does response validation do anything here? See: https://github.com/kentik/community_sdk_golang/pull/16#discussion_r591681147

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: Added a follow-up question

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: Done

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: Done

Fenthick (changes requested):
Couple comments added

kentikapi/internal/api_resources/saved_filters_test.go
lukaa12: Ok I will add TODO

kentikapi/internal/api_payloads/saved_filter.go
lukaa12: I think it's unnecessary here and I will remove it

kentikapi/internal/api_payloads/saved_filter.go
```diff
@@ -0,0 +1,194 @@
...
+	savedFilterPayload
+}
+
+type UpdateSavedFilterResponse struct {
+	savedFilterPayload
+}
+
+type savedFilterPayload struct {
+	ID                IntAsString    `json:"id,omitempty" response:"get,post,put"`
+	CompanyID         models.ID      `json:"company_id,string,omitempty" response:"get,post,put"`
```
Fenthick, line 43: Please take a look: https://github.com/kentik/community_sdk_golang/pull/16/files#r595079608

Fenthick (changes requested):

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: I see that the link to comment doesn't work somehow. I linked following comment:

> I see that ID and Company ID still have validation tags:
> 
> ```
> ID                IntAsString    `json:"id,omitempty" response:"get,post,put"`
> CompanyID         models.ID      `json:"company_id,string,omitempty" response:"get,post,put"`
> ```
> 
> Aren't they redundant?

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: It's a pity I cannot reopen the conversation BTW

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: The validation is only for pointer, interface or slice (newly added):

validate.go:55
`if (field.Kind() == reflect.Ptr || field.Kind() == reflect.Interface || field.Kind() == reflect.Slice) && field.IsNil()`

So it does nothing for IntAsString and models.ID.

What's the purpose? I don't fully know. I used response tags to avoid nil-pointer dereference when I used `time.Time` in model and `*time.Time` in payload.

SNMPSpeed is IntAsString so request validation tag does nothing and is IMO redundant. Maybe the intention was to return error also on zero value of a field (so `0` for int), but this is not how current implementation works.

So I propose to remove those tags and move on.

mateuszmidor: Tags allow to validate if the request/response payload has all the required attributes set to not-nil, so the user gets better error message than "Invalid request" from the API or than segfault when dereferencing nil in response payload. So it is mainly for pointer/interface types. Rarely for slices as slice can be assigned nil and API doesnt like this in some cases

lukaa12: Ok, then I will remove these tags from savedFilterPayload's ID, CompanyID and InterfacePayload SNMPSpeed

Fenthick (approved):

kentikapi/internal/api_payloads/saved_filter.go
Fenthick: Resolved

mateuszmidor (approved):

TedatKentik (approved):
Saved filters implementation

PR #15, opened 2021-03-03
Implement Users create/delete methods

Fenthick:
Issue: KNTK-161

PR #17, opened 2021-03-04
Implement Users update method

Fenthick:
Issue: KNTK-161

This PR is based on https://github.com/kentik/community_sdk_golang/pull/15


kentikapi/models/user.go
```diff
@@ -2,24 +2,45 @@ package models
...
-	UserEmail    string     `json:"user_email"`
-	Role         string     `json:"role"`
-	EmailService bool       `json:"email_service"`
-	EmailProduct bool       `json:"email_product"`
-	LastLogin    *time.Time `json:"last_login"`
-	CreatedDate  time.Time  `json:"created_date"`
-	UpdatedDate  time.Time  `json:"updated_date"`
-	CompanyID    ID         `json:"company_id,string"`
+// NewUser creates a new User with all required fields set.
+func NewUser(u UserRequiredFields) *User {
```
Fenthick, line 37: https://kb.kentik.com/v0/Ec06.htm#Ec06-User_Create says that all those fields are required.
In fact, only UserEmail is required by the API, rest are optional.
In Python library we treat password and role as optional. 

I implemented it to behave just like API documentation. @mateuszmidor @lukaa12 What do you think?

BTW I removed `user_password` field, because it seems that there is no such thing in the API.

mateuszmidor (approved):

kentikapi/models/user.go
```diff
@@ -2,24 +2,45 @@ package models
 
 import "time"
 
-type GetAllUsersResponse struct {
-	Users []User `json:"users"`
+// User model.
+type User struct {
+	// read-write properties (can be updated in update call)
+	Username     string
```
Fenthick, line 8: I just checked with Postman, that updating username returns 200, but does not update that field.
I think that I should mark this field as not updatable (read-only), what do you think?

lukaa12: I agree, if it cannot be updated then it should be marked properly.

kentikapi/models/user.go
Fenthick: Stayed with all required choice.

mateuszmidor (approved):

lukaa12 (approved):

TedatKentik (approved):
implement user update methods

PR #31, opened 2021-03-31
Disable response enums validation for Device API

Fenthick:
There are no more resources with response enums left.
See: ADR #3

Issue: KNTK-189


lukaa12 (approved):

mateuszmidor (approved):

TedatKentik (approved):
Enums validation - more elegant fail mode, when unexpected responses arrive

PR #43, opened 2021-07-09
Regenerate the APIv6 SDK and fix examples

Fenthick:
Regenerate API v6 client and stub server code.

Improve generate* scripts:
1. Pin openapi-generator version
2. Download latest OpenAPI schema
3. Gofmt generated code
4. Generate cloud export server code to "cloudexport" directory instead
   of "cloud_export"
5. Remove redundant files/directories
6. Add manually modified files to .openapi-generator-ignore

Fix compilation and runtime of API v6 examples.

Issue: KNTK-283


apiv6/generate_client.sh
```diff
@@ -1,52 +1,79 @@
...
         exit 1
     fi
 
     echo "Done"
 }
 
-function generate_golang_client_from_openapi3_spec() {
-    stage "Generating golang client from openapi spec"
+function download_openapi_spec() {
+    wget "$cloud_export_spec_url" --output-document "$cloud_export_spec_filename"
```
mmac-m3a, line 32: Won't work if user does not have `wget` (I for one don't have it). I suggest to use `curl` which is more common and more versatile.

apiv6/generate_client.sh
```diff
@@ -1,52 +1,79 @@
...
-
     echo "Done"
 }
 
 function change_ownership_to_current_user() {
     dir="$1"
     stage "Changing ownership of $dir to $USER:$USER"
 
-    sudo chown  -R "$USER:$USER"  "$dir" # by default the generated output is in user:group root:root
+    sudo chown -R "$USER:$USER" "$dir" # by default the generated output is in user:group root:root
```
mmac-m3a, line 76: Does not work in my environment, because there is no group with the same name as my username. MacOS does not create a separate group for each user. I suggest to just use `chown -R "$USER" ...`.

BTW, This has been failing for me for a while, but it does not seem to have any negative effect. Maybe the chown is even not needed.

apiv6/generate_client.sh
```diff
@@ -56,33 +83,24 @@ function cleanup_non_needed_files() {
...
-change_ownership_to_current_user "$cloudexport_client_output_dir"
-cleanup_non_needed_files "$cloudexport_client_output_dir"
-
-# GENERATE SYNTHETICS
-synthetics_package="synthetics"
-synthetics_spec="synthetics.openapi.yaml"
+function stage() {
+    BOLD_BLUE="\e[1m\e[34m"
+    RESET="\e[0m"
+    msg="$1"
```
mmac-m3a, line 100: The color tricks do not always work in my environment. Most of the time, it just spews control codes.

mmac-m3a (commented):
I cannot say I've really looked at all of the 96 modified files, but other than the comments regarding usage of `wget` and assumptions about user groups, what I've seen looked good.

apiv6/generate_client.sh
Fenthick: Switched to `curl` and added it to prerequisites

apiv6/generate_client.sh
Fenthick: Modified to not change the group. Changing user is needed so that generated files can be removed without `sudo` and formatted by `gofmt`.

apiv6/generate_client.sh
Fenthick: Created bug: https://clkentik.atlassian.net/browse/KNTK-288

apiv6/generate_client.sh
```diff
@@ -56,33 +92,29 @@ function cleanup_non_needed_files() {
 
     generated_content_dir="$1"
 
+    rm "$generated_content_dir/.travis.yml"
+    rm "$generated_content_dir/git_push.sh"
     rm "$generated_content_dir/go.mod"
     rm "$generated_content_dir/go.sum"
     rm "$generated_content_dir/README.md"
-    rm -rf "$generated_content_dir/docs"
     rm -rf "$generated_content_dir/api"
```
mateuszmidor, line 100: If the spec is autodownloaded, please keep the source spec around for reference if something breaks

mateuszmidor (approved):

apiv6/generate_client.sh
Fenthick: Modified the script not to remove the spec and docs. Docs do not break anything and maybe might be useful for quickly glancing the functionality by library devs. All gitignored.

mmac-m3a (approved):

PR #45, opened 2021-07-21
Modify localhost_apiserver not to require preset in GET /tests

Fenthick:
Issue: KNTK-270

### DIFF
--- a/apiv6/generate_server.sh
+++ b/apiv6/generate_server.sh
@@ -15,6 +15,7 @@ function run() {
     generate_cloud_export_server
     generate_synthetics_server
     go fmt ./...
+    git apply ./localhost_apiserver/patches/0001-Modify-localhost_apiserver-not-to-require-preset-in-.patch
 }
 
 function check_prerequisites() {
--- a/apiv6/go.mod
+++ b/apiv6/go.mod
@@ -3,7 +3,6 @@ module github.com/kentik/community_sdk_golang/apiv6
 go 1.15
 
 require (
-	github.com/gorilla/handlers v1.5.1
 	github.com/gorilla/mux v1.8.0
 	github.com/hashicorp/go-retryablehttp v0.7.0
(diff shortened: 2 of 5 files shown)


apiv6/generate_server.sh
```diff
@@ -15,6 +15,7 @@ function run() {
     generate_cloud_export_server
     generate_synthetics_server
     go fmt ./...
+    git apply ./localhost_apiserver/patches/0001-Modify-localhost_apiserver-not-to-require-preset-in-.patch
```
ziemowit141, line 18: Why is application of this patch required, when you also actually changed the: apiv6/localhost_apiserver/synthetics/go/api_synthetics_admin_service.go

Does it make any change it that way?

apiv6/go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/kentik/community_sdk_golang/apiv6
 go 1.15
 
 require (
-	github.com/gorilla/handlers v1.5.1
```
ziemowit141, line 6: Just out of curiosity, is this change related to the ticket? I do not see removing usage of this package in those changes.

apiv6/generate_server.sh
Fenthick: This way, every time the schema changes, the file will be regenerated accordingly. After regeneration is done, the patch with the fix is applied. I chose this way in order to keep the file up to date with the latest schema.

Other solution would be to disable generation of this file by adding it to `.openapi-generator-ignore`.

apiv6/go.mod
Fenthick: No, I have just run `go mod tidy`

ziemowit141 (approved):
All my concerns got addressed

mmac-m3a (approved):
Fine with me, since I hope to convince Kentik engineering to remove this parameter entirely.

PR #46, opened 2021-08-16
Move API v5 client code to apiv5 directory. Issue: KNTK-193

Opelord:
apiv5 files now in proper directory
added bash test script in tools directory


Fenthick:
Please add reference to the Jira issue in commit description and PR description.
It could look like: `Issue: KNTK-193`

Fenthick:
Commit summary should be short. I suggest to use `Move API v5 client code to apiv5 directory`.
I think that PR name should be more informative. I suggest to use the same as the commit summary: `Move API v5 client code to apiv5 directory`

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
Fenthick, line 1: Please, do not commit IDE-related files. So please remove `.idea` directory
I suggest to use `git commit -u` instead of `git commit -A` to avoid checking in unintended files.

.github/workflows/test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Checkout code
```
Fenthick, line 25: Please do not move LICENSE file. It belongs to the repository, not particular library. (I can't put comment on LICENSE file)

Fenthick (changes requested):

apiv5/README.md
```diff
@@ -24,7 +24,11 @@ export KTAPI_AUTH_EMAIL=<your kentik api credentials email>
 export KTAPI_AUTH_TOKEN=<your kentik api credentials token>
 
 # Run from the repository root
-go test -tags examples -count 1 -v ./examples/users_example_test.go
+./tools/test.sh -v apiv5 -t TestDemonstrateUsersGetAll
```
Fenthick, line 27: As discussed offline:
- please pass all positional arguments as test args passed to `go test` calls
- do not run examples on CI
- there is no need for -v and -t options if we allow user to pass test args

PR #56, opened 2021-09-21
Added support for RFC1123 Date in Retry-After header

Opelord:
Issue: KNTK-276
Handle Retry-After timestamp header Go API v6 client

Go-retryablehttp package doesn't support Retry-After header in form of a timestamp. [This issue](https://github.com/hashicorp/go-retryablehttp/issues/144) has been created alongside a few PRs to add this functionality but for now I have forked their package and updated it with [this PR](https://github.com/hashicorp/go-retryablehttp/pull/138). 
Hopefully in near future we will receive developer's comment on the issue and the changes to go.mod can be reverted.


apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
 	}
 }
 
+func TestRetryAfter(t *testing.T) {
```
Fenthick, line 110:
```suggestion
func TestRetryingClient_Do_HandlesRetryAfterHeader(t *testing.T) {
```
To keep test naming consistent and to make the name informative.

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
...
+	var (
+		header     string
+		httpErrors = []int{
+			http.StatusBadGateway,
+			http.StatusTooManyRequests,
+			http.StatusTooManyRequests,
+			http.StatusBadRequest,
+		}
+		retryAfterValue = 3 * time.Second
```
Fenthick, line 119: This could be const, I think.

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
 	}
 }
 
+func TestRetryAfter(t *testing.T) {
+	var (
+		header     string
```
Fenthick, line 112: I think that this variable is not needed here.

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// act
+	//nolint:bodyclose
+	resp, err := c.Do(req.WithContext(context.Background()))
+
+	// assert
+	t.Logf("Got response: %v, err: %v", resp, err)
+
+	assert.Equal(t, resp.StatusCode, 400)
```
Fenthick, line 148:
```suggestion
	assert.Equal(t, 400, resp.StatusCode)
```
> func Equal(t TestingT, expected, actual interface{}, msgAndArgs ...interface{}) bool {

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
...
+	assert.Equal(t, resp.Header["Retry-After"][0], time.Now().Add(retryAfterValue).Format(time.RFC1123))
```
Fenthick, line 149: I think that this check is flaky, because `time.Now()` here is a little bit later than in test server.
I doubt that we need to check that client propagates correct header value.

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,48 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
...
+	})
+
+	req, err := retryablehttp.NewRequest(http.MethodGet, ts.URL, nil)
+	require.NoError(t, err)
+
+	// act
+	//nolint:bodyclose
+	resp, err := c.Do(req.WithContext(context.Background()))
+
+	// assert
```
Fenthick, line 145: This test does not verify that duration of retry-after is used by retry mechanism.
I propose following modifications:

1. httpErrors to contain only retry-after-compatible statuses, except last one: [StatusServiceUnavailable, StatusTooManyRequests, StatusOk]
2. Set both MinDelay and MaxDelay to large numbers, e.g. 10 Seconds.
3. Set context.Timeout passed to c.Do() to e.g. 3 Seconds.
4. Set retryAfterValue to low value e.g. 10 Milliseconds.
5. Verify that there is no error returned from c.Do() and no error in context (there was no timeout).

apiv6/examples/demos/retryafter_demo/main.go
```diff
@@ -0,0 +1,191 @@
+package main
```
Fenthick, line 1: I suggest to adjust retry_demo so that latest retry-after functionality is showcased there to avoid excess duplications. I wonder why dupl didn't catch that.

Fenthick (changes requested):

apiv6/examples/demos/retryafter_demo/main.go
Opelord: I wondered this too. I think `run: skip-dirs` option in linters' config file is skipping more directories than intended. I will take a look at it.

apiv6/kentikapi/httputil/httputil_test.go
Opelord: The problem with low value for retryAfterValue is the format of RFC1123. `RFC1123     = "Mon, 02 Jan 2006 15:04:05 MST"`
This only takes into account the seconds part. The way this works when I run the same test with e.g. retry-after header 500 ms is it sometimes retries after random amount of time between 0ms and 250-ish ms but majority of time it's 0ms because time.Now() + 500 ms is approximated to full second and usually already in the past when server reads when to retry. 
I suggest 1 second for retry-after or - if that's too long - complete restructuring of the test.

Fenthick: Ok, you can use 1 second for retryAfterValue then.
Using 10 ms wouldn't be too bad either, because it would still test that the code respects the retry-after header, which is the most important thing here.

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: I still think that `TestRetryingClient` is better prefix that `TestNewRetryingClient`, because we test functionality of retrying client, not only its constructor.

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

apiv6/kentikapi/httputil/httputil_test.go
```diff
@@ -106,6 +107,47 @@ func TestRetryingClientWithSpyHTTPTransport_Do(t *testing.T) {
...
+	defer cancel()
+
+	//nolint:bodyclose
+	resp, err := c.Do(req.WithContext(ctx))
+
+	// assert
+	t.Logf("Got response: %v, err: %v", resp, err)
+
+	assert.NoError(t, err)
+	assert.Equal(t, 200, resp.StatusCode)
```
Fenthick, line 148:
```suggestion
	assert.Equal(t, http.StatusOK, resp.StatusCode)
```
Please use the same form as in test input (missed that earlier).

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

apiv6/examples/demos/retryafter_demo/main.go
Fenthick: Done

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

mmac-m3a (approved):

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

apiv6/kentikapi/httputil/httputil_test.go
Fenthick: Done

Fenthick (approved):

mmac-m3a (approved):

PR #63, opened 2021-11-03
Transform API v6 client to use gRPC

Opelord:
Issue: KNTK-256
In this PR we implement the basic features of gRPC client. It is capable of sending a request and handling a response while using metadata headers as authetication methods.
I suggest reviewing each commit separately.

PR #65, opened 2021-11-19
Implement retry strategy for gRPC client

Opelord:
Issue: KNTK-357
This PR is an extension of PR #63 which was closed and moved to PR #66


kentikapi/client.go
```diff
@@ -228,3 +260,23 @@ func (c Config) makeAuthInterceptor() grpc.UnaryClientInterceptor {
 		return invoker(ctx, method, req, reply, cc, opts...)
 	}
 }
+
+func (c Config) makeRetryInterceptor() grpc.UnaryClientInterceptor {
+	opts := []grpc_retry.CallOption{
```
Fenthick, line 265: I am pretty sure that there is no need to create local slice - you can just pass those 3 options to directly to _UnaryClientInterceptor()_ call

kentikapi/client.go
```diff
@@ -228,3 +260,23 @@ func (c Config) makeAuthInterceptor() grpc.UnaryClientInterceptor {
...
+		grpc_retry.WithBackoff(grpc_retry.BackoffExponential(*c.RetryCfg.MinDelay)),
+		grpc_retry.WithCodes(codes.Unavailable),
+		// TODO: ensure c.RetryCfg.MaxAttempts is positive or research other options (mpalczynski)
```
Fenthick, line 268: Let's just change type of c.RetryCfg.MaxAttempts to _uint_, so that the compiler enforces sane values.

kentikapi/client_integration_test.go
```diff
@@ -478,11 +434,90 @@ func TestClient_GetAgent(t *testing.T) {
...
+			responses: []gRPCResponse{
+				newErrorGRPCResponse(codes.Unavailable),
+				newErrorGRPCResponse(codes.Unavailable),
+				{
+					nil,
+					&syntheticspb.GetAgentResponse{Agent: newDummyAgentForGRPC()},
+				},
+			},
+			expectedResult:    &syntheticspb.GetAgentResponse{Agent: newDummyAgentForGRPC()},
+			expectedErrorCode: codes.OK,
```
Fenthick, line 450: This is the case when error is nil, so its code and msg are not checked. So those 2 lines are misleading, because our API does not return error code at all. Please drop those 2 lines

kentikapi/client_integration_test.go
```diff
@@ -492,14 +527,21 @@ func TestClient_GetAgent(t *testing.T) {
 				AuthToken:              dummyAuthToken,
 				AuthEmail:              dummyAuthEmail,
 				DisableTLS:             true,
+				RetryCfg: kentikapi.RetryConfig{
+					MaxAttempts: tt.retryMax,
+					MinDelay:    durationPtr(10 * time.Millisecond),
```
Fenthick, line 532: MinDelay could be in microseconds, so that test time is minimized.

kentikapi/client_integration_test.go
```diff
@@ -518,7 +560,7 @@ func TestClient_GetAgent(t *testing.T) {
 				assert.Equal(t, tt.expectedRequestID, r.GetId())
 			}
 
-			assert.Equal(t, tt.expectedResult.GetAgent(), response.GetAgent())
+			assert.Equal(t, tt.expectedResult.GetAgent().String(), response.GetAgent().String())
```
Fenthick, line 563: What was the problem with comparing objects?
When there is difference in string representation, there will be no precise diff displayed, so it is inferior solution.

kentikapi/internal/httputil/httputil.go
```diff
@@ -106,9 +86,12 @@ func defaultHTTPClient() *http.Client {
 // makeRetryPolicy creates customized retry policy.
 // Its implementation is based on retryablehttp.ErrorPropagatedRetryPolicy.
 // Retry policy function returns true if the request should be retried.
-func makeRetryPolicy(statusCodes []int, methods []string) retryablehttp.CheckRetry {
-	statusCodesSet := makeIntSet(statusCodes)
-	methodsSet := makeStringSet(methods)
+func makeRetryPolicy() retryablehttp.CheckRetry {
+	statusCodesSet := makeIntSet([]int{429, 500, 502, 503, 504})
```
Fenthick, line 90: I think that we can remove status 500 from defaults. community_sdk_python HTTP library doesn't return on 500 and Go gRPC part of code does not return on such "unknown" server errors.

kentikapi/internal/httputil/httputil.go
```diff
@@ -106,9 +86,12 @@ func defaultHTTPClient() *http.Client {
...
+	methodsSet := makeStringSet([]string{
```
Fenthick, line 91: How about removing methodsSet and `if !isRequestRetryable(resp.Request, methodsSet) {` block of code? It is not needed now.

Fenthick (changes requested):

kentikapi/client.go
```diff
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"time"
 
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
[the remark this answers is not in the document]
mmac-m3a: Unless the run-on unreadable words (like `grpcsynthetics`) are some kind of "standard" in the go-land, I'd prefer to use some delimiters.

examples/demos/interfaces_query_demo/main.go
```diff
@@ -13,7 +14,10 @@ import (
 const deviceID = models.ID(80603) // crated-before demo router
```
mmac-m3a: Typo in comment (I'm pretty sure that "crated" should be "created"). Also, why not created the `device` as part of the demo.

mmac-m3a (commented):
This code also should not go to master until the use of gRPC transport is complete and invisible to SDK users.

kentikapi/client.go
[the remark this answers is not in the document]
palczyn: Going by established packages in Go, like `jsoniter` or `retryablehttp` I think the change Daniel suggested is the proper way in Golang.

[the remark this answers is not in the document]
danielfurman: Yes, the idiomatic Go way (TM) is to melt everything into one word. But it might be cargo-culting this sentence:
> By convention, packages are given lower case, single-word names; there should be no need for underscores or mixedCaps.
https://go.dev/doc/effective_go#package-names

kentikapi/internal/httputil/httputil.go
```diff
@@ -61,28 +57,12 @@ type RetryConfig struct {
...
-	RetryableStatusCodes []int
-	// RetryableMethods are HTTP request retry methods, which the retry strategy is enabled for.
-	// Default: [GET, HEAD, POST, PUT, PATCH, DELETE, CONNECT, OPTIONS, TRACE].
-	RetryableMethods []string
 }
 
 func (cfg *ClientConfig) FillDefaults() {
 	if cfg.HTTPClient == nil {
 		cfg.HTTPClient = defaultHTTPClient()
 	}
```
danielfurman, line 76: Please call cfg.RetryConfig.FillDefaults() here so that httputil.NewRetryingClient() has safe interface and does not panic when empty config is passed

kentikapi/client.go
```diff
@@ -162,6 +151,44 @@ func NewClient(c Config) (*Client, error) {
...
+
+	if c.SyntheticsGRPCHostPort == "" {
+		c.SyntheticsGRPCHostPort = syntheticsGRPCHostPort
+	}
+
+	if c.Timeout == nil {
+		c.Timeout = durationPtr(defaultTimeout)
+	}
+
+	if c.RetryCfg.MaxAttempts == nil {
```
danielfurman: I think that it could be extracted to c.RetryConfig.FillDefaults() that would be called here. That way we can also use that method in httputil.

kentikapi/internal/httputil/httputil.go
```diff
@@ -25,16 +25,12 @@ func NewRetryingClient(cfg ClientConfig) *retryablehttp.Client {
...
-		c.RetryMax = *cfg.RetryCfg.MaxAttempts
-	}
-	if cfg.RetryCfg.MinDelay != nil {
-		c.RetryWaitMin = *cfg.RetryCfg.MinDelay
-	}
-	if cfg.RetryCfg.MaxDelay != nil {
-		c.RetryWaitMax = *cfg.RetryCfg.MaxDelay
-	}
-	c.CheckRetry = makeRetryPolicy(cfg.RetryCfg.RetryableStatusCodes, cfg.RetryCfg.RetryableMethods)
+
```
danielfurman: I would sleep better if you wouldn't remove that nil checks. There is always risk that pointers are uninitialized, as the code evolves

Fenthick (changes requested):

kentikapi/client_integration_test.go
palczyn: On the two occasions that the SpyServer returns valid result this field is different: `state: impl.MessageState{ atomicMessageInfo:(*impl.MessageInfo)(nil)}}`. For some reason when I try to use t.Logf just before this assert the tests pass. It's really weird.

palczyn: `String()` method assures that I'm only comparing the data that's interesting for users and not data associated with a given message.

> // The MessageState type holds a *MessageInfo, which must be atomically set to
// the message info associated with a given message instance.

examples/demos/interfaces_query_demo/main.go
Opelord: I don't know, Martin. I was not in Codilime when it was written.

kentikapi/client.go
palczyn: But they work a bit differently, because httputil.go fills the fields RetryMax, RetryWaitMin and RetryWaitMax of the retryablehttp client and client.go fills the fields in the configuration.
I've made some adjustments let me know after the next commit if it's a good approach.

Opelord:
Actually I don't know how to move the whole discussion to another PR so I will keep this one open and after the changes are approved I will rebase and PR to the proper branch.

kentikapi/internal/httputil/httputil.go
danielfurman: Please move `cfg.RetryCfg.FillDefaults()` call to the body of `cfg.FillDefaults()`  (i.e. `ClientConfig.FillFefaults()`). That way filling defaults is in one place (at the beggining of `NewRetryingClient()`) and is before initializations.

Fenthick (changes requested):

kentikapi/client.go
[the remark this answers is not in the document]

kentikapi/client.go

kentikapi/client.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go
danielfurman: Please use proto.Equal for comparing proto objects.

See: https://github.com/stretchr/testify/issues/758#issuecomment-645282184

kentikapi/internal/httputil/httputil.go
danielfurman: Done

kentikapi/internal/httputil/httputil.go
danielfurman: Done

examples/demos/interfaces_query_demo/main.go
danielfurman: First sentence done.

kentikapi/client.go
danielfurman: Your adjustment is almost what I wanted. I left one small comment in comment above.

kentikapi/internal/httputil/httputil.go
danielfurman: Done

examples/demos/interfaces_query_demo/main.go
```diff
@@ -40,7 +45,22 @@ func main() {
 	queryChart(client)
 }
 
-func createInterface(client *kentikapi.Client) models.ID {
+func createDevice(client *kentikapi.Client) models.ID {
+	device := models.NewDeviceDNS(
+		"testapi_for_demo",
```
danielfurman:
```suggestion
		"interfaces_query_demo_device",
```
To avoid conflicts and help with manual cleanup, if needed

examples/demos/interfaces_query_demo/main.go
```diff
@@ -40,7 +45,22 @@ func main() {
...
+		models.DeviceSubtypeAwsSubnet,
+		1,
+		11466,
+		models.CDNAttributeYes,
+	)
+	createdDevice, err := client.Devices.Create(context.Background(), *device)
+	demos.ExitOnError(err)
+	fmt.Printf("Successfuly created interface, ID = %d\n", createdDevice.ID)
```
danielfurman:
```suggestion
	fmt.Printf("Successfuly created device, ID = %d\n", createdDevice.ID)
```

kentikapi/client_integration_test.go
```diff
@@ -558,7 +559,7 @@ func TestClient_GetAgent(t *testing.T) {
 				assert.Equal(t, tt.expectedRequestID, r.GetId())
 			}
 
-			assert.Equal(t, tt.expectedResult.GetAgent().String(), response.GetAgent().String())
+			assert.True(t, proto.Equal(tt.expectedResult.GetAgent(), response.GetAgent()))
```
danielfurman:
```suggestion
			assert.True(t, proto.Equal(tt.expectedResult, result), fmt.Sprintf("Protobuf messages are not equal:\nexpected: %v\nactual:  %v", tt.expectedResult, result))
```
Notice: please rename response variable _result_

kentikapi/internal/httputil/httputil.go
danielfurman: Done

Fenthick (changes requested):

kentikapi/client_integration_test.go
danielfurman: Done; I suggested adding msg in separate comment

examples/demos/interfaces_query_demo/main.go
danielfurman: Second sentence done as well

kentikapi/client.go

examples/demos/interfaces_query_demo/main.go
danielfurman: Done

examples/demos/interfaces_query_demo/main.go
danielfurman: Done

kentikapi/client_integration_test.go

Fenthick (approved):

Fenthick (approved):

examples/demos/interfaces_query_demo/main.go
```diff
@@ -3,30 +3,39 @@ package main
...
 	demos.Step("Get all interfaces")
-	getAllInterfaces(client)
+	getAllInterfaces(client, deviceID)
+
+	// Delete Device
+	demos.Step("Delete device")
+	deleteDevice(client, deviceID)
 
 	// Query
 	demos.Step("Query for data")
```
mmac-m3a, line 41: I wonder what querying flow data has to do with demonstrating interface related operations.

kentikapi/client.go
```diff
@@ -1,27 +1,41 @@
...
 const (
-	APIURLUS          = "https://api.kentik.com/api/v5"
-	APIURLEU          = "https://api.kentik.eu/api/v5"
-	cloudExportAPIURL = "https://cloudexports.api.kentik.com"
-	syntheticsAPIURL  = "https://synthetics.api.kentik.com"
+	APIURLUS                = "https://api.kentik.com/api/v5"
+	APIURLEU                = "https://api.kentik.eu/api/v5"
+	cloudExportAPIURL       = "https://cloudexports.api.kentik.com"
+	syntheticsAPIURL        = "https://synthetics.api.kentik.com"
+	syntheticsGRPCHostPort  = "synthetics.api.kentik.com:443"
```
mmac-m3a, line 37: We could parse/infer the host and port of the gRPC API from the `syntheticsAPIURL`. Due to the implementation of the service, it is 99.99% certain that the host and port will _always_ be the same for both HTTP/REST and gRPC APIs. It is safe to assume that if the `protocol` part of the URL is `https` (and the URL does not contain any explicit `port` specification) that the port is 443.

kentikapi/client.go
```diff
@@ -39,11 +53,20 @@ type Client struct {
 	Plans              *resources.PlansAPI
 	Alerting           *resources.AlertingAPI
 
+	// CloudExportAdminServiceAPI, SyntheticsAdminServiceAPI and SyntheticsDataServiceAPI are http clients
```
mmac-m3a, line 56: Since we now have the code on a dev branch, simply remove the HTTP/REST code for all APIs that have been already converted to gRPC. The goal is to convert _all_ existing v6 APIs to gRPC and verify that external behavior is identical to their HTTP/REST counterparts _before_ the `dev_grpc` branch is merged to `master`. Ultimately the change should be entirely transparent to users. They will simply start using gRPC once they rebuild their app with new version of the SDK (Terraform providers will be the guinea pigs :)).

kentikapi/client.go
```diff
@@ -55,37 +78,46 @@ type Config struct {
...
-	AuthToken        string
-	RetryCfg         RetryConfig
+	// SyntheticsGRPCHostPort defaults to "synthetics.api.kentik.com:443".
+	SyntheticsGRPCHostPort string
+	// CloudExportGRPCHostPort defaults to "cloudexport.api.kentik.com:443".
+	CloudExportGRPCHostPort string
+	AuthEmail               string
+	AuthToken               string
+	// RetryCfg has no effect on CloudExportAdmin, SyntheticsAdmin and SyntheticsData.
+	RetryCfg RetryConfig
```
mmac-m3a, line 88: Gee, I thought that implementing retry strategy is the main point of the PR :-/.

kentikapi/client_integration_test.go
```diff
@@ -5,32 +5,39 @@ import (
...
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 const (
 	testAgentID = "968"
 )
 
-func TestClient_PatchAgent(t *testing.T) {
+func TestClient_PatchAgentHTTP(t *testing.T) {
```
mmac-m3a, line 32: I strongly prefer that the integration test uses only gRPC. I may be incomplete/failing for some time, but eventually it should work exactly as it's HTTP/REST counterpart. Ideally, we should have test data (stored in JSON or some other suitable format) that are used both by the HTTP/REST and gRPC fake servers, so that both HTTP and gRPC clients using them are guaranteed to get the same responses for equivalent requests. Do you see any problem with this approach?

examples/demos/retry_apiv5_demo/main.go
```diff
@@ -41,16 +41,17 @@ func showRetryingOnMultipleCodes() error {
...
 	demos.Step("Create Kentik API v5 client")
-	c := kentikapi.NewClient(kentikapi.Config{
+	c, err := kentikapi.NewClient(kentikapi.Config{
 		APIURL: s.URL,
 		RetryCfg: kentikapi.RetryConfig{
-			MaxAttempts:          intPtr(42),
+			MaxAttempts:          uintPtr(42),
 			MinDelay:             durationPtr(1 * time.Second),
 			MaxDelay:             durationPtr(2 * time.Second),
-			RetryableStatusCodes: []int{http.StatusTooManyRequests, http.StatusBadGateway, http.StatusServiceUnavailable},
```
mmac-m3a, line 50: I hope we have not lost the ability to specify retryable status codes/methods. It is entirely unclear to me from the code :-/.

mmac-m3a (approved):
I've not noticed any implementation of configurable retry strategy for gRPC client in this PR. Where have I missed it?
I strongly prefer to remove all HTTP code for APIs that are already converted to gRPC. Keeping both makes the code hard to read and leads to unnecessary naming compromises.

examples/demos/retry_apiv5_demo/main.go
palczyn: We decided to simplify code for gRPC since all of these status codes in HTTP are condensed to `codes.Unavailable`. We can try to implement our own set of codes that will be translated to gRPC and HTTP if you need this feature.

kentikapi/client.go
danielfurman: cloudExportAPIURL and syntheticsAPIURL will be removed with HTTP API v6 code, so there will be no duplication.

We could tough construct syntheticsGRPCHostPort from APIURLUS, but that would hard-code `synthetics.` subdomain, which we should probably avoid.

kentikapi/client.go
danielfurman: Removal tracked in https://clkentik.atlassian.net/browse/KNTK-370

kentikapi/client.go
danielfurman: @Opelord This comment needs to be removed

palczyn: Yes, it will be changed in the next PR

examples/demos/interfaces_query_demo/main.go
palczyn: To be fair, this file has been created on 5th of March, I only updated it. Maybe it was needed then.

PR #67, opened 2021-11-25
Added examples of gRPC for API v6

palczyn:
Issue: KNTK-366
First commit contains examples of Synthetics API and second Synthetics Mesh Tests and Cloud Export API.

PR #68, opened 2021-12-01
Implemented timeout for gRPC client

palczyn:
Issue: KNTK-360
This PR is build on PR #67


examples/synthetics_example_test.go
```diff
@@ -51,20 +55,22 @@ func runAdminServiceExamples() error {
...
+	if err != nil {
+		return err
+	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), 300*time.Second)
 	defer cancel()
 
-	// NOTE: test of id 3336 must exist
-	if err = runGetHealthForTests(ctx, client, []string{"3336"}); err != nil {
+	// NOTE: test of id 3541 must exist
```
mmac-m3a, line 66: Why not list tests and pick and existing one?

examples/synthetics_example_test.go
```diff
@@ -258,27 +264,27 @@ func runCRUDAgent(ctx context.Context, client *kentikapi.Client) error {
...
-	//PrettyPrint(patchResp)
-	//fmt.Println()
+	fmt.Println("### PATCH AGENT")
+	agent := *getResp.Agent
+	if getResp.Agent.GetFamily() == synthetics.V202101BETA1IPFAMILY_V6 {
+		agent.SetFamily(synthetics.V202101BETA1IPFAMILY_V4)
+	} else {
+		agent.SetFamily(synthetics.V202101BETA1IPFAMILY_V6)
+	}
+	agent.Name = nil
```
mmac-m3a, line 274: I think you may also need to set `site` in the payload, even if it is not being changed (it is a bug in validation). In any case, don't put much effort in PATCH requests. We will remove support for PATCH in the "stable" version of the API and instead support PUT for updating existing instance.

examples/synthetics_example_test.go
```diff
@@ -258,27 +264,27 @@ func runCRUDAgent(ctx context.Context, client *kentikapi.Client) error {
...
+		AgentPatch(context.Background(), agentID).
+		Body(patchReqPayload).
+		Execute()
+	if err != nil {
+		return fmt.Errorf("%v %v", err, httpResp)
+	}
+	PrettyPrint(patchResp)
+	fmt.Println()
 
 	// NOTE: as we can't create agents through the API - let's not delete them
```
mmac-m3a, line 289: Reasonable choice, however agents actually can be deleted via the API.

examples/synthetics_example_test.go
```diff
@@ -258,27 +264,27 @@ func runCRUDAgent(ctx context.Context, client *kentikapi.Client) error {
...
-	//	Body(patchReqPayload).
-	//	Execute()
-	//if err != nil {
-	//	return fmt.Errorf("%v %v", err, httpResp)
-	//}
-	//PrettyPrint(patchResp)
-	//fmt.Println()
+	fmt.Println("### PATCH AGENT")
+	agent := *getResp.Agent
+	if getResp.Agent.GetFamily() == synthetics.V202101BETA1IPFAMILY_V6 {
```
mmac-m3a, line 269: A better example would be changing status of agent from AGENT_STATUS_WAIT to AGENT_STATUS_OK, which some customers already do as part of automated deployment of private agents. You can use the `agent_activate` method in [synth_tools](https://github.com/kentik/synth_tools/blob/f9cb686312a90b0439e97f3496d7bf31614d4e45/synth_tools/commands/agents.py#L94) as a example.

examples/synthetics_example_test.go
```diff
@@ -321,7 +327,7 @@ func runListAgents(ctx context.Context, client *kentikapi.Client) error {
 // prepare a Test for sending in CREATE request
 func makeExampleTest() *synthetics.V202101beta1Test {
 	hostname := synthetics.NewV202101beta1HostnameTest()
-	hostname.SetTarget("dummy-ht")
+	hostname.SetTarget("dummy-ht.com")
```
mmac-m3a, line 330: `www.example.com` is a good name to use.

examples/synthetics_example_test.go
```diff
@@ -392,3 +398,375 @@ func makeExampleTest() *synthetics.V202101beta1Test {
...
+	}
+
+	getHealthResp, err := client.SyntheticsData.GetHealthForTests(ctx, healthPayLoad)
+	if err != nil {
+		return err
+	}
+
+	if getHealthResp.Health != nil {
+		healthItems := getHealthResp.GetHealth()
+		fmt.Println("Num health items:", len(healthItems))
```
mmac-m3a, line 565: "Num health items will be always equal to number entries in `testIDs` array (assuming that those tests exist). The returned data is a list  of entries one per requested test ID.

examples/synthetics_example_test.go
```diff
@@ -392,3 +398,375 @@ func makeExampleTest() *synthetics.V202101beta1Test {
...
+
+	fmt.Println()
+
+	return nil
+}
+
+func runGRPCCRUDAgent(ctx context.Context, client *kentikapi.Client) error {
+	// NOTE: no CREATE method exists for agents in the API, thus no example for CREATE
+	// NOTE: agent of id 1717 must exist
+	agentID := "1717"
```
mmac-m3a, line 621: Again, why not list agents and pick an existing one?

examples/synthetics_example_test.go
```diff
@@ -392,3 +398,375 @@ func makeExampleTest() *synthetics.V202101beta1Test {
...
+
+	test := &syntheticspb.Test{
+		Name:      "example-test-1",
+		Type:      "hostname",
+		DeviceId:  "1000",
+		Status:    syntheticspb.TestStatus_TEST_STATUS_ACTIVE,
+		Settings:  settings,
+		ExpiresOn: timestamppb.New(time.Now().Add(time.Hour * 6)),
+		Cdate:     timestamppb.Now(),
+		CreatedBy: userInfo,
```
mmac-m3a, line 768: I apologize for lack of documentation on meaning and acceptable values for attributes in the API. I still own the team a document (shame on me). @Fenthick knows quite a bit about it, because he has tested it pretty thoroughly in the Terraform synthetics agent.
In any case:
- `device_id` is an internal attribute that will be soon removed from the API. It should be hidden from users and set to 0 for now
- `cdate`, `edate` and `created_by` are read-only (and populated by the backend)
- expires_on` is non-functional and shall be ignored for now. `status` is ignored on create (and always set to `TEST_STATUS_ACTIVE`).

examples/synthetics_example_test.go
```diff
@@ -392,3 +398,375 @@ func makeExampleTest() *synthetics.V202101beta1Test {
...
+		Ping:               ping,
+		Trace:              trace,
+		Port:               443,
+		Protocol:           "icmp",
+		Family:             syntheticspb.IPFamily_IP_FAMILY_DUAL,
+		Servers:            []string{},
+		UseLocalIp:         false,
+		Reciprocal:         false,
+		RollupLevel:        1,
+		Http:               nil,
```
mmac-m3a, line 751: `hostname` type test will never have any `http` section. In general, each test type except `application_mesh` and `bgp_monitor` has a section `settings` named after the test type (i.e. for `hostname` test, the name is `hostname`). This section contains test type specific attributes, most importantly the `target` (or `targets`) and eventually more. There are some anomalies (e.g. `dns` and `dns_grid` tests require list of `servers` and it is not in the test type specific section, rather directly in `settings`). Most (if not all) of those anomalies are going to be straightened in `beta2` version of the API (which already exists, but has not been made public yet). This is though just a stepping stone to "stable" version which will have additional schema changes.

examples/synthetics_mesh_example_test.go
```diff
@@ -13,9 +13,10 @@ import (
...
 
-	"github.com/stretchr/testify/assert"
-
+	syntheticspb "github.com/kentik/api-schema-public/gen/go/kentik/synthetics/v202101beta1"
 	"github.com/kentik/community_sdk_golang/kentikapi/synthetics"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 var testID = flag.String("testid", "3541", "id of mesh test to display the result matrix for")
```
mmac-m3a, line 22: Again, list tests and pick one with type `application-mesh`.

examples/synthetics_mesh_example_test.go
```diff
@@ -152,3 +156,131 @@ func (m metricsMatrix) getMetric(fromAgent string, toAgent string) (*synthetics.
...
+		header = header + x + "\t"
+	}
+	fmt.Fprintln(w, header)
+
+	// print table rows
+	for _, fromAgent := range matrix.agents {
+		row := fromAgent + "\t"
+		for _, toAgent := range matrix.agents {
+			if metrics, ok := matrix.getMetricGRPC(fromAgent, toAgent); ok {
+				row = row + formatLatencyGRPC(metrics) + "\t"
```
mmac-m3a, line 225: Why not also packet loss and jitter?

examples/synthetics_mesh_example_test.go
```diff
@@ -152,3 +156,131 @@ func (m metricsMatrix) getMetric(fromAgent string, toAgent string) (*synthetics.
...
+	client, err := NewClient()
+	if err != nil {
+		return nil, err
+	}
+
+	healthPayload := &syntheticspb.GetHealthForTestsRequest{
+		Ids:       []string{testID},
+		StartTime: timestamppb.New(time.Now().Add(-time.Minute * 5)),
+		EndTime:   timestamppb.Now(),
+		Augment:   true, // if not set, returned Mesh pointer will be empty
```
mmac-m3a, line 188: Despite that, all of the data you may want to display will be present (just not it the `mesh` section) :-).

kentikapi/client.go
```diff
@@ -150,3 +213,61 @@ func makeRetryingClientConfig(c Config) httputil.ClientConfig {
...
+		ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs(
+			authEmailKey, c.AuthEmail,
+			authAPITokenKey, c.AuthToken,
+		))
+		return invoker(ctx, method, req, reply, cc, opts...)
+	}
+}
+
+func (c Config) makeRetryInterceptor() grpc.UnaryClientInterceptor {
+	return grpcretry.UnaryClientInterceptor(
```
mmac-m3a, line 263: Good, here is some code handling retries!

mmac-m3a (approved):
Same general comment as for PR#65 - I'd prefer to remove the HTTP/REST related code in order to avoid unnecessary naming collisions, etc.

I'll merge this PR to keep moving forward.

kentikapi/client_integration_test.go
```diff
@@ -401,6 +401,8 @@ func TestClient_GetAgent(t *testing.T) {
 		expectedErrorCode codes.Code
 		expectedErrorMsg  string
 		expectedError     bool
+		timeout           *time.Duration
```
danielfurman, line 404: Timeout is config argument so put it just below retryMax

kentikapi/client_integration_test.go
```diff
@@ -401,6 +401,8 @@ func TestClient_GetAgent(t *testing.T) {
...
+		responseDelay     *time.Duration
```
danielfurman, line 405: responseDelay is option for spy server, so put it next to `responses` field

kentikapi/client_integration_test.go
```diff
@@ -511,13 +513,37 @@ func TestClient_GetAgent(t *testing.T) {
...
 			expectedError:     true,
+		}, {
+			name:              "timeout after 100ms",
+			request:           &syntheticspb.GetAgentRequest{},
+			responses:         []gRPCResponse{},
+			expectedResult:    nil,
+			expectedErrorCode: codes.DeadlineExceeded,
+			expectedErrorMsg:  "context deadline exceeded",
+			expectedError:     true,
+			timeout:           durationPtr(100 * time.Millisecond),
```
danielfurman, line 524: Please reduce it to 1 ms max to make test quicker

kentikapi/client_integration_test.go
```diff
@@ -511,13 +513,37 @@ func TestClient_GetAgent(t *testing.T) {
...
+			timeout:           durationPtr(125 * time.Millisecond),
+			responseDelay:     durationPtr(50 * time.Millisecond),
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			// arrange
 			s := newSpySyntheticsServer(t, tt.responses)
+			s.handlingDelay = tt.responseDelay
```
danielfurman, line 546: Please pick and use one name

kentikapi/client_integration_test.go
```diff
@@ -579,6 +606,9 @@ type spySyntheticsServer struct {
 	// requests spied by the server
 	requests []*syntheticspb.GetAgentRequest
 	headers  []metadata.MD
+
+	// handlingDelay specifies the delay applied while handling the request
+	handlingDelay *time.Duration
```
danielfurman, line 611: Nit: Please put it next to responses, because it is connected with it - the way the server handles the request.

BTW: It does not need to be a pointer, because sleep returns immediately with zero value duration

kentikapi/client_integration_test.go
```diff
@@ -614,6 +644,11 @@ func (s *spySyntheticsServer) GetAgent(ctx context.Context, req *syntheticspb.Ge
 ) (*syntheticspb.GetAgentResponse, error) {
 	header, ok := metadata.FromIncomingContext(ctx)
 	assert.True(s.t, ok)
+
+	if s.handlingDelay != nil {
```
danielfurman, line 648: Please put it before or after spy code, not in the middle of that.

examples/synthetics_example_test.go
palczyn: How about being able to set testID when running this example from command window?

danielfurman: It's better to pick existing test programmatically than manually

PR #72, opened 2021-12-22
Simplified test and function names in API v6 examples

palczyn:
Issue: KNTK-256

PR #73, opened 2021-12-27
Set optional parameters via pointers package

palczyn:
Issue: KNTK-195
Based on PR #72


examples/query_example_test.go
```diff
@@ -83,9 +84,9 @@ func runQueryChart() error {
 
 	fmt.Println("### QUERY Chart")
 	agg1 := models.NewAggregate("avg_bits_per_sec", "f_sum_both_bytes", models.AggregateFunctionTypeAverage)
```
palczyn, line 84: I don't get why we used `NewAggregate()` here and in the demo we used `Aggregate{}`. Should I make it more consistent?

danielfurman: New*() are constructors that take required arguments. Using them is encouraged. I don't know if there is a specific reason not to use them in demo.

palczyn: Sorry I meant `kentikapi/internal/resources/query_test.go`, not demo.

danielfurman: I suggest to use the constructor in tests as well. So that they are closer the real usage

danielfurman (approved):

mateuszmidor (approved):

examples/demos/client_merge_demo/main.go
```diff
@@ -47,7 +50,7 @@ func showClientMerge() error {
 	getResp, err := c.SyntheticsAdmin.ListAgents(context.Background(), &syntheticspb.ListAgentsRequest{})
```
mmac-m3a: Indentation of the line above seems to be of.

examples/demos/client_merge_demo/main.go
```diff
@@ -47,7 +50,7 @@ func showClientMerge() error {
...
 
 	fmt.Println("Received result:")
-	demos.PrettyPrint(getResp)
+	demos.PrettyPrint(getResp.GetAgents())
```
mmac-m3a: It would be good to show difference (or absence of difference) between results received via v5 and v6 API.

examples/demos/devices_users_demo/main.go
```diff
@@ -42,12 +45,12 @@ func createDevice(client *kentikapi.Client) models.ID {
 		models.ID(11466),
 		models.CDNAttributeYes,
 	)
-	models.SetOptional(&device.SiteID, 8483)
-	models.SetOptional(&device.DeviceBGPFlowSpec, true)
+	device.SiteID = models.IDPtr(8483)
+	device.DeviceBGPFlowSpec = pointer.ToBool(true)
```
mmac-m3a, line 49: Cannot the value be directly assigned? Most reasonable programming languages represent bool as some atomic type.

mmac-m3a (approved):
I trust the author that the change is worth the hassle. I don't quite see it ... but there more things that I do not understand about Golang ;-).

mmac-m3a:
Please, rebase the PR and I'll merge it.

examples/demos/client_merge_demo/main.go
danielfurman: Good indentation is ensured by CI. The issue is that GH UI displays tabs inconsistently.

examples/demos/client_merge_demo/main.go
danielfurman: `result` is a slice of Users and `getResp.GetAgents()` is a slice of Agents. Those are entirely different objects.

examples/demos/devices_users_demo/main.go
danielfurman: We use pointers to achieve optional values. `DeviceBGPFlowSpec` field has `*bool` type. Pointer to bool must be assigned here. Such literals as "true" are not addressable, so it cannot be assigned inline. In other words, you cant do `device.DeviceBGPFlowSpec = &true`.

See: https://stackoverflow.com/a/30716481/5000820

danielfurman:
@mmac-m3a Task description contains rationale for this change: https://clkentik.atlassian.net/browse/KNTK-195

examples/demos/client_merge_demo/main.go
mmac-m3a: Well, I probably then do not understand what this example demonstrates ... but that does not need to hold up the merge ;).

PR #74, opened 2021-12-29
Optimization of example cases in repository

palczyn:
Based on PR #73 
- Added pickTenantID() function to avoid stating tenant ID explicitly in use cases (KNTK-373)
- Added t.Parallel() to shorten execution time (146s to 83s on my pc). (KNTK-339)
- Fixed a bug were goroutine that ran the gRPC spy server could fail after the test has been completed. (KNTK-374)


kentikapi/client_integration_test.go
```diff
@@ -220,6 +221,7 @@ type spySyntheticsServer struct {
 	syntheticspb.UnimplementedSyntheticsAdminServiceServer
 	server *grpc.Server
 	url    string
+	done   chan bool
```
danielfurman: Please use `struct{}` type as token sent through channel

kentikapi/client_integration_test.go
```diff
@@ -166,7 +166,6 @@ func TestClient_GetAgent(t *testing.T) {
 			s := newSpySyntheticsServer(t, tt.responses)
 			s.handlingDelay = tt.handlingDelay
 			s.Start()
-			defer s.Stop()
```
danielfurman: Please keep `defer s.Stop()` here to ensure resources are always freed.

kentikapi/client_integration_test.go
```diff
@@ -212,6 +211,8 @@ func TestClient_GetAgent(t *testing.T) {
 				proto.Equal(tt.expectedResult, result),
 				fmt.Sprintf("Protobuf messages are not equal:\nexpected: %v\nactual:  %v", tt.expectedResult, result),
 			)
+			s.Stop()
+			assert.True(t, <-s.done)
```
danielfurman: I think that it should be moved to `(s *spySyntheticsServer) Stop() {}` function, so that the the user has only to call one method to stop the server.

kentikapi/client_integration_test.go
```diff
@@ -252,12 +255,14 @@ func (s *spySyntheticsServer) Start() {
...
-		require.NoError(s.t, s.server.Serve(l))
+		err = s.server.Serve(l)
+		require.NoError(s.t, err)
+		s.done <- true
 	}()
 }
 
 func (s *spySyntheticsServer) Stop() {
-	s.server.GracefulStop()
+	s.server.Stop()
```
danielfurman, line 265: What were the issues with GracefulStop?

danielfurman (changes requested):

kentikapi/client_integration_test.go
palczyn: I don't really understand what you mean by this.

kentikapi/client_integration_test.go
palczyn: Good idea

kentikapi/client_integration_test.go
palczyn: Stop always makes tests one second shorter because because it doesn't wait for the last response in the last test to return like GracefulStop() does. Since we don't need it I changed it to Stop().

kentikapi/client_integration_test.go
palczyn: Ok, I think I found out what you meant.

mateuszmidor (approved):

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

danielfurman (approved):

examples/cloud_export_example_test.go
```diff
@@ -14,11 +15,12 @@ import (
 )
 
 func TestCloudExportAPIExample(t *testing.T) {
-	assert.NoError(t, runGRPCCRUDCloudExport())
-	assert.NoError(t, runGRPCGetAllCloudExports())
+	t.Parallel()
```
mmac-m3a, line 18: Why do we want to run _examples_ in parallel (or more effectively in any way)? I'd personally prefer to keep things serial, so that the recipient of the demo can understand it.

mmac-m3a (commented):
KNTK-374 and KNTK-373 make sense to me.
I don't see much reason to make examples run faster. KNTK-339 does not make much sense to me. Can you, please, explain the rationale to me.
It is PITA to go over (pretty much) the same 42 files that I've already seen in #73. I'd rather not have to do it.

examples/cloud_export_example_test.go
danielfurman: To speed development time.
`t.Parallel()` allows examples to be run in parallel, but providing `-parallel 1` to `go test` will run them sequentially. I added such flag to README.

mmac-m3a (approved):

PR #78, opened 2022-01-28
Get credentials pub func

jdziurzynski:
KNTK-371
Add functionality to read credentials from env variables or file, similarly as in python library.

### DIFF
--- a/kentikapi/credentials.go
+++ b/kentikapi/credentials.go
@@ -1,0 +1,93 @@
+package kentikapi
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+)
+
+const DefaultProfile = "default"
+
+// GetCredentials reads and returns (email, token) pair from environment variables or specified profile config file.
+// Profile name specifies file name containing Kentik profile data. DefaultProfile constant can be used for that.
+// Credentials lookup order (assuming profile name is "default"):
+// 1. env vars: KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN
+// 2. file at $KTAPI_CFG_FILE
+// 3. file at $KTAPI_HOME/.kentik/default
+// 4. file at $HOME/.kentik/default
+// 5. file at ./.kentik/default
+// Profile file format: {"email":"dummy@acme.com","api-key":"dummy"}.
+func GetCredentials(profileName string) (authEmail, authToken string, err error) {
+	authEmail, authToken = readCredentialsFromEnv()
+	if authEmail == "" || authToken == "" {
+		authEmail, authToken, err = readCredentialsFromProfile(profileName)
+	} else {
+		log.Printf("Credentials loaded from environment variables\n")
+	}
+	return authEmail, authToken, err
+}
+
+// ReadCredentialsFromEnv reads and returns (email, token) pair from environment variables.
+func readCredentialsFromEnv() (authEmail, authToken string) {
+	authEmail, _ = os.LookupEnv("KTAPI_AUTH_EMAIL")
+	authToken, _ = os.LookupEnv("KTAPI_AUTH_TOKEN")
+	return authEmail, authToken
+}
+
+type profile struct {
+	AuthEmail string `json:"email"`
+	AuthToken string `json:"api-key"`
+}
+
+// ReadCredentialsFromProfile reads and returns (email, token) pair from profile config file.
+func readCredentialsFromProfile(profileName string) (authEmail, authToken string, err error) {
+	var result profile
+	filename := getFilename(profileName)
+	jsonFile, err := os.Open(filepath.Clean(filename))
+	if err != nil {
+		return "", "", err
+	}
+	byteValue, err := ioutil.ReadAll(jsonFile)
+	if err != nil {
+		return "", "", err
+	}
+	if cErr := jsonFile.Close(); cErr != nil {
+		return "", "", cErr
+	}
+
+	err = json.Unmarshal(byteValue, &result)
+	if err != nil {
+		return "", "", err
+	}
+	if result.AuthEmail == "" {
+		return "", "", fmt.Errorf("no AuthEmail in config file: %s", filename)
+	}
+	if result.AuthToken == "" {
+		return "", "", fmt.Errorf("no AuthToken in config file: %s", filename)
+	}
+	log.Printf("Credentials loaded from %v\n", filename)
+	return result.AuthEmail, result.AuthToken, nil
+}
+
+func getFilename(profileName string) string {
+	home := resolveHome()
+	filename, ok := os.LookupEnv("KTAPI_CFG_FILE")
+	if !ok || filename == "" {
+		return filepath.Join(home, ".kentik", profileName)
+	}
+	return filename
+}
+
+func resolveHome() string {
+	home, ok := os.LookupEnv("KTAPI_HOME")
+	if !ok || home == "" {
+		home, ok = os.LookupEnv("HOME")
+		if !ok || home == "" {
+			home = "."
+		}
+	}
+	return home
+}
--- a/examples/utils.go
+++ b/examples/utils.go
@@ -2,32 +2,15 @@
 package examples
 
 import (
-	"errors"
 	"fmt"
-	"os"
 	"reflect"
 
 	"github.com/kentik/community_sdk_golang/kentikapi"
 )
 
-// ReadCredentialsFromEnv reads and returns (email, token) pair from environment variables, or error if not set.
-func ReadCredentialsFromEnv() (authEmail, authToken string, _ error) {
-	authEmail, ok := os.LookupEnv("KTAPI_AUTH_EMAIL")
-	if !ok || authEmail == "" {
-		return "", "", errors.New("KTAPI_AUTH_EMAIL environment variable needs to be set")
-	}
-
-	authToken, ok = os.LookupEnv("KTAPI_AUTH_TOKEN")
-	if !ok || authToken == "" {
-		return "", "", errors.New("KTAPI_AUTH_TOKEN environment variable needs to be set")
-	}
-
-	return authEmail, authToken, nil
-}
-
-// NewClient creates Kentik API client with credentials read from environment variables.
+// NewClient creates kentikapi client with credentials read from env variables.
 func NewClient() (*kentikapi.Client, error) {
-	email, token, err := ReadCredentialsFromEnv()
+	email, token, err := kentikapi.GetCredentials(kentikapi.DefaultProfile)
 	if err != nil {
 		return nil, err
 	}


kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	authToken, _ = os.LookupEnv("KTAPI_AUTH_TOKEN")
+	return authEmail, authToken
+}
+
+// ReadCredentialsFromProfile reads and returns (email, token) pair from profile config file.
+func ReadCredentialsFromProfile() (authEmail, authToken string, err error) {
+	var result map[string]string
+	profile := "default"
+	filename := getFilename(profile)
+	jsonFile, err := os.Open(filepath.Clean(filename))
```
mateuszmidor, line 49: Why filepath.Clean here...?

danielfurman:
Please add issue number in PR/commit descriptions.

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	filename, ok := os.LookupEnv("KTAPI_CFG_FILE")
+	if !ok || filename == "" {
+		return filepath.Join(home, ".kentik", profile)
+	}
+	return filename
+}
+
+// GetCredentials reads and returns (email, token) pair from environment variables or profile config file.
+func GetCredentials() (authEmail, authToken string, err error) {
```
danielfurman: Supporting smaller APIs is easier, so I would keep only this function exported. Additionally, please move it to the top o a file.

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	authEmail, authToken = ReadCredentialsFromEnv()
+	if authEmail == "" || authToken == "" {
+		authEmail, authToken, err = ReadCredentialsFromProfile()
+		if authEmail == "" || authToken == "" || err != nil {
+			err = fmt.Errorf("%w\nKTAPI_AUTH_EMAIL and KTAPI_AUTH_TOKEN variables needs to be set in environment or "+
+				"should be provided in file (./.kentik/default example: {\"email\": \"john.doe@domain.com\",\"api-key\": \"12345\"})", err)
```
danielfurman: Please run `tools/fmt.sh`

danielfurman (changes requested):

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+	filename, ok := os.LookupEnv("KTAPI_CFG_FILE")
+	if !ok || filename == "" {
+		return filepath.Join(home, ".kentik", profile)
+	}
+	return filename
+}
+
+// GetCredentials reads and returns (email, token) pair from environment variables or profile config file.
```
danielfurman:
```suggestion
// GetCredentials reads and returns (email, token) pair from environment variables or specified profile config file.
// Environment variables used: KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN
// Default profile name is "default".
// Profile config file path is ${HOME}/.kentik/default by default. It can be overridden by KTAPI_CFG_FILE env variable or KTAPI_HOME variable used to path "${KTAPI_HOME}/.kentik/<profile_name>".
// Profile file format: <TODO>
```
Let's have usage documentation here, so that user doesn't need to read code to use the function.

Edit: removed KTAPI_PROFILE env

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+func ReadCredentialsFromEnv() (authEmail, authToken string) {
+	authEmail, _ = os.LookupEnv("KTAPI_AUTH_EMAIL")
+	authToken, _ = os.LookupEnv("KTAPI_AUTH_TOKEN")
+	return authEmail, authToken
+}
+
+// ReadCredentialsFromProfile reads and returns (email, token) pair from profile config file.
+func ReadCredentialsFromProfile() (authEmail, authToken string, err error) {
+	var result map[string]string
+	profile := "default"
```
danielfurman, line 21: I think that user should have ability to override it with KTAPI_PROFILE env variable

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	for _, s := range []string{"email", "api-key"} {
+		if _, ok := result[s]; !ok {
+			return "", "", fmt.Errorf("no '%s' in config file: %s", s, filename)
+		}
+	}
+	return result["email"], result["api-key"], nil
+}
+
+func getFilename(profile string) string {
+	home, ok := os.LookupEnv("KTAPI_HOME")
```
danielfurman: Please extract resolveHome() function

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	var result map[string]string
+	profile := "default"
+	filename := getFilename(profile)
+	jsonFile, err := os.Open(filepath.Clean(filename))
+
+	if err != nil {
+		return "", "", err
+	}
+	defer func() {
+		cerr := jsonFile.Close()
```
danielfurman:
```suggestion
		cErr := jsonFile.Close()
```

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+// ReadCredentialsFromEnv reads and returns (email, token) pair from environment variables.
+func ReadCredentialsFromEnv() (authEmail, authToken string) {
+	authEmail, _ = os.LookupEnv("KTAPI_AUTH_EMAIL")
+	authToken, _ = os.LookupEnv("KTAPI_AUTH_TOKEN")
+	return authEmail, authToken
+}
+
+// ReadCredentialsFromProfile reads and returns (email, token) pair from profile config file.
+func ReadCredentialsFromProfile() (authEmail, authToken string, err error) {
+	var result map[string]string
```
danielfurman: Please unmarshall to struct defined as:
```go
type profile struct {
	AuthEmail string `json:"email"`
	AuthToken string `json:"api-key"`
}
```
This is a good practice in Go programming.

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	return filename
+}
+
+// GetCredentials reads and returns (email, token) pair from environment variables or profile config file.
+func GetCredentials() (authEmail, authToken string, err error) {
+	authEmail, authToken = ReadCredentialsFromEnv()
+	if authEmail == "" || authToken == "" {
+		authEmail, authToken, err = ReadCredentialsFromProfile()
+		if authEmail == "" || authToken == "" || err != nil {
```
danielfurman: Please modify ReadCredentialsFromProfile to check if email and token are not empty and return error otherwise.
Check only error message here. That way we get more precise error message.

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+	if !ok || filename == "" {
+		return filepath.Join(home, ".kentik", profile)
+	}
+	return filename
+}
+
+// GetCredentials reads and returns (email, token) pair from environment variables or profile config file.
+func GetCredentials() (authEmail, authToken string, err error) {
+	authEmail, authToken = ReadCredentialsFromEnv()
+	if authEmail == "" || authToken == "" {
```
danielfurman:
```suggestion
	if authEmail != "" && authToken != "" {
	   return authEmail, authToken, nil
	}
```

I suggest to flip the logic to avoid indentation.
Similar to this idea:  https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

kentikapi/credentials.go
```diff
@@ -0,0 +1,78 @@
...
+		authEmail, authToken, err = ReadCredentialsFromProfile()
+		if authEmail == "" || authToken == "" || err != nil {
+			err = fmt.Errorf("%w\nKTAPI_AUTH_EMAIL and KTAPI_AUTH_TOKEN variables needs to be set in environment or "+
```
danielfurman: I think error message could be a little bit simpler and we should provide precise details in docstring.
Something like: ("no credentials in [KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN] environment variables nor Kentik profile file: %v", err)

kentikapi/credentials.go
danielfurman:
```suggestion
const DefaultProfile = "default"

func GetCredentials(profile string) (authEmail, authToken string, err error) {
```

Python library allows user to pick profile and we should match that functionality. Globally exported DefaultProfile const could be passed by users in typical situation.

danielfurman (changes requested):

kentikapi/credentials.go
jdziurzynski: Using filepath.Clean() to clean up possible bad paths. 
Without filepath.Clean() linter returning given error: G304: Potential file inclusion via variable (gosec)

mateuszmidor: OK so this is linter thing, OS would resolve the path anyway

examples/custom_applications_example_test.go
```diff
@@ -15,6 +15,8 @@ import (
 )
 
 func TestCustomApplicationsAPIExample(t *testing.T) {
+	t.Skipf("Kentik API is broken")
```
mateuszmidor: Error message from Kentik: `API response error, status: 500 Internal Server Error, response body: {"error":"API unexpected issue"}` is misleading; the actual problem is that a Custom Application named "apitest-customapp-1" already exists and its ID is 614. Must have been created and not deleted afterwards - this happens when an example crashes after creating a resource, for one or another reason. 
Please remove it manually eg. here https://api.kentik.com/api/tester/v5#custom_application-DELETE-Custom-Application-delete and try running the example again

mateuszmidor:
After testing the function I recommend adding logging at debug level the fact, that the final source of the credentials was either:

1. env vars: KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN
2. file at $KTAPI_CFG_FILE
3. file at $KTAPI_HOME/.kentik/<profile name>
4. file at $HOME/.kentik/<profile name>
5. file at ./.kentik/<profile name>

It may be confusing at times, which credentials were actually loaded

kentikapi/credentials.go
```diff
@@ -0,0 +1,88 @@
...
+	AuthToken string `json:"api-key"`
+}
+
+const DefaultProfile = "default"
+
+// GetCredentials reads and returns (email, token) pair from environment variables or specified profile config file.
+// Environment variables used: KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN.
+// Default profile name can be specified by global variable DefaultProfile.
+// Profile config file path is ${HOME}/.kentik/default by default. It can be overridden by
+// KTAPI_CFG_FILE env variable or KTAPI_HOME variable used to path "${KTAPI_HOME}/.kentik/<profile_name>".
```
mateuszmidor: Would be nice to have a comment like this:

Credentials lookup order (assuming profile name is "default"):
1. env vars: KTAPI_AUTH_EMAIL, KTAPI_AUTH_TOKEN
2. file at $KTAPI_CFG_FILE
3. file at $KTAPI_HOME/.kentik/default
4. file at $HOME/.kentik/default
5. file at ./.kentik/default

kentikapi/credentials.go
mateuszmidor: It's helpful to have the explicit data structure, I agree.

mateuszmidor (approved):

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Still to do

kentikapi/credentials.go
```diff
@@ -40,39 +64,32 @@ func ReadCredentialsFromProfile() (authEmail, authToken string, err error) {
...
-	return result["email"], result["api-key"], nil
+	if result.AuthToken == "" {
+		return "", "", fmt.Errorf("no AuthToken in config file: %s", filename)
+	}
+	log.Printf("Credentials loaded from %v\n", filename)
+	return result.AuthEmail, result.AuthToken, nil
 }
 
-func getFilename(profile string) string {
+func resolveHome() string {
```
danielfurman, line 84: We have a practice in Go codebase that helper functions are below their usages. So please move resolveHome below getFilename

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
```diff
@@ -4,31 +4,55 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 )
 
+type Profile struct {
```
danielfurman: This type is an implementation detail, not meant to be consumed by users so please unexport it (i.e. make private) and move next to readCredentialsFromProfile() function. I would put it below readCredentialsFromProfile function.

kentikapi/credentials.go
```diff
@@ -4,31 +4,55 @@ import (
...
+	AuthEmail string `json:"email"`
+	AuthToken string `json:"api-key"`
+}
+
+const DefaultProfile = "default"
+
+// Credentials lookup order (assuming profile name is "default"):
```
danielfurman, line 16: Please put it into GetCredentials docstring, so that it is rendered into pkg.go.dev.

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: I still the logic should be flipped, but is is minor stylistic thing.

kentikapi/credentials.go
danielfurman: Error message simplified. Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Still to do

kentikapi/credentials.go
```diff
@@ -24,10 +19,11 @@ const DefaultProfile = "default"
 // 5. file at ./.kentik/default
 
 // GetCredentials reads and returns (email, token) pair from environment variables or specified profile config file.
-func GetCredentials(profile string) (authEmail, authToken string, err error) {
+// Default profileName name can be specified by global variable DefaultProfile.
```
danielfurman:
```suggestion
// Profile name specifies file name containing Kentik profile data. DefaultProfile constant can be used for that.
```
It's not variable, it's a const.

danielfurman (changes requested):

kentikapi/credentials.go
danielfurman: @jdziurzynski Still to do. Order of credentials sources given by Mateusz and profile file format will be enough

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: @jdziurzynski Please add profile file format (e.g. example JSON) to the docstring.

kentikapi/credentials.go
danielfurman: Done

kentikapi/credentials.go
danielfurman: Done

danielfurman:
@jdziurzynski Profile file format in docstring stil to be added

kentikapi/credentials.go
danielfurman: Done

danielfurman (approved):

mmac-m3a (approved):

mmac-m3a:
Please, resolve conflicts and check failures.

mmac-m3a (approved):

PR #79, opened 2022-02-01
Bash scripts colors methods refactored

jdziurzynski:
Issue: KNTK-288
Color codes used in Kentik open source repositories are broken on MacOS. 
Changed stage() methods to use same colored_echo() method.

### DIFF
--- a/examples/run_example.sh
+++ b/examples/run_example.sh
@@ -1,17 +1,18 @@
 #!/usr/bin/env bash
 
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+REPO_DIR=$(cd -- "$SCRIPT_DIR" && cd ../ && pwd)
+
+source "$REPO_DIR/tools/utility_functions.sh" || exit 1
+
 # INPUT:
 #   [none] - run all examples
 #   User
 #   Device
 
 function stage() {
-    BLUE_BOLD="\e[1m\e[34m"
-    RESET="\e[0m"
-    msg="$1"
-
     echo
-    echo -e "$BLUE_BOLD$msg$RESET"
+    colored_echo BOLD_BLUE "$1"
 }
 
 if [[ $1 != "" ]]; then
--- a/examples/run_example.sh
+++ b/examples/run_example.sh
@@ -1,24 +1,14 @@
 #!/usr/bin/env bash
 
-SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
-REPO_DIR=$(cd -- "$SCRIPT_DIR" && cd ../ && pwd)
-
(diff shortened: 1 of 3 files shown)


examples/run_example.sh
```diff
@@ -1,17 +1,18 @@
 #!/usr/bin/env bash
```
danielfurman: IMO for such simple script we should use echo instead of stage and remove all that jazz.
BTW this script is meant to be executed from `examples` directory. Otherwise it will spam logs from all unit/integration tests.

kentikapi/gen_enum.sh
```diff
@@ -1,12 +1,12 @@
...
-    RESET="\e[0m"
-    msg="$1"
+SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
+REPO_DIR=$(cd -- "$SCRIPT_DIR" && cd ../ && pwd)
+
+source "$REPO_DIR/tools/utility_functions.sh" || exit 1
 
-    echo
-    echo -e "$COLOR$msg$RESET"
+function stage() {
```
danielfurman: Please just use `stage` function from utility functions

danielfurman (changes requested):

examples/run_example.sh
danielfurman: Done

kentikapi/gen_enum.sh
danielfurman: Done

danielfurman (approved):

mateuszmidor (approved):

examples/run_example.sh
```diff
@@ -5,19 +5,10 @@
 #   User
 #   Device
```
mateuszmidor, line 6: Maybe it's worth to actually include all available examples here?
It's nowhere documented how user can derive the name.

danielfurman: The place for that is in readme in "running examples" section.
BTW, I consider this script redundant, because one line needed to run an example is provided in readme.

mmac-m3a (approved):

examples/run_example.sh
```diff
@@ -5,19 +5,10 @@
...
-    RESET="\e[0m"
-    msg="$1"
-
-    echo
-    echo -e "$BLUE_BOLD$msg$RESET"
-}
-
 if [[ $1 != "" ]]; then
-    stage "Running single example: $1"
+    echo "Running single example: $1"
```
mmac-m3a, line 9: Shouldn't the `colored_echo` function be used here? It surely does not matter much, but this is mentioned as the goal of the change in the PR decription.

jdziurzynski: In previous review I got suggestion that coloring this certain one line is not required and to simplify this file as possible, and so I did.

PR #80, opened 2022-02-01
Refactor examples and Client

danielfurman:
Refactors extracted from API v6 transformation
Remove context.WithTimeout from examples, because API v6 supports
client-level timeout now, so context timeout is not needed.

Issue: KNTK-372


mateuszmidor (approved):
TestQueryAPIExample timed out once while testing, but seems to be server-side cause.

mmac-m3a (approved):
LGTM

PR #85, opened 2022-03-16
Fix payloads apiv6

jdziurzynski:
KNTK-379

Added new interceptor to enable logging payloads.


kentikapi/client.go
```diff
@@ -156,6 +156,7 @@ func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error
 				makeTimeoutInterceptor(c),
 				makeAuthInterceptor(c),
 				makeRetryInterceptor(c),
+				makeLoggerInterceptor(c),
```
danielfurman: Interceptors are executed left to right, so logger should be higher. I suggest to make it the first interceptor, so that logging the request (and response) is always performed.

See: https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware@v1.3.0#hdr-Chaining

kentikapi/client.go
```diff
@@ -200,3 +201,21 @@ func makeRetryInterceptor(c Config) grpc.UnaryClientInterceptor {
 		grpcretry.WithMax(*c.RetryCfg.MaxAttempts+1),
 	)
 }
+
+func makeLoggerInterceptor(c Config) grpc.UnaryClientInterceptor {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn,
+		invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
```
danielfurman, line 207: Nit: I suggest to break the line after last function parameter, just like in makeTimeoutInterceptor() to increase readability

danielfurman: Nit: I suggest to break the line after last function parameter, just like in makeTimeoutInterceptor() to increase readability

kentikapi/client.go
```diff
@@ -200,3 +201,21 @@ func makeRetryInterceptor(c Config) grpc.UnaryClientInterceptor {
...
+		c.LogPayloads = true
```
danielfurman, line 208: Excuse me what the puck?

danielfurman: Excuse me what the puck?

kentikapi/client.go
```diff
@@ -200,3 +201,21 @@ func makeRetryInterceptor(c Config) grpc.UnaryClientInterceptor {
...
+		if c.LogPayloads {
+			log.Printf("Kentik API request - %s - %+v\n", method, req)
+		}
+		err := invoker(ctx, method, req, reply, cc, opts...)
+		if err != nil {
```
danielfurman, line 213: No. Logging response should not be skipped when underlying code returns error. We need that information in logs!

danielfurman: No. Logging response should not be skipped when underlying code returns error. We need that information in logs!

kentikapi/client.go
```diff
@@ -200,3 +201,21 @@ func makeRetryInterceptor(c Config) grpc.UnaryClientInterceptor {
...
+			return err
+		}
+		if c.LogPayloads {
+			log.Printf("Kentik API response - %s - %+v n", method, reply)
```
danielfurman, line 217:
```suggestion
			log.Printf("Kentik API response - %s - %+v n", method, reply)
```
```suggestion
			log.Printf("Kentik API response - %s - %+v\n", method, reply)
```

Please be more diligent

danielfurman:
```suggestion
			log.Printf("Kentik API response - %s - %+v\n", method, reply)
```

danielfurman (changes requested):
Please be more diligent. I suggest to always read your own PRs before requesting review

danielfurman (changes requested):

kentikapi/client.go
mateuszmidor: LogPayloads should not be modified here I guess?

kentikapi/client.go
mateuszmidor: Not sure, what do we want to log here? The logging interceptor simply dumps the method+payload, it seems reasonable for it to be close to the actual request invocation so eg. every retried request is logged?

mmac-m3a: I'd prefer to log the request once and then log messages indicating retry (which should not contain full payload, just some unique identifier of the request being retried.

kentikapi/client.go
mmac-m3a: Absolutely!

danielfurman:
@jdziurzynski I am seeing huge diff right now. Please rebase the PR with squash

kentikapi/client.go
danielfurman: Still to do. I agree with Martin. I think that logging interceptor should be above auth interceptor. No strong opinion on position relative to timeout interceptor.

kentikapi/client.go
danielfurman: Please just remove this whole `if err != nil {}` block of code.

danielfurman (changes requested):

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

danielfurman (approved):

mmac-m3a (approved):

PR #86, opened 2022-03-17
Simplified url configuration

jdziurzynski:
KNTK-382

Url configuration simplified to a single URL

Full task description:
https://docs.google.com/document/d/1b-huL-uHSVyI3eaao3-YOFHOZ115Sz9FOjlNOdHIULY/edit?pli=1#heading=h.bilt7bxurkl6

### DIFF
--- a/kentikapi/client.go
+++ b/kentikapi/client.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"net"
+	"net/url"
 	"time"
 
 	"github.com/AlekSi/pointer"
@@ -29,10 +31,8 @@ const (
 
 // Kentik API URLs.
 const (
-	APIURLUS            = "https://api.kentik.com/api/v5"
-	APIURLEU            = "https://api.kentik.eu/api/v5"
-	syntheticsHostPort  = "synthetics.api.kentik.com:443"
-	cloudExportHostPort = "cloudexport.api.kentik.com:443"
+	APIURLUS = "https://api.kentik.com"
+	APIURLEU = "https://api.kentik.eu"
 )
 
 // Client is the root object for manipulating all the Kentik API resources.
@@ -61,15 +61,12 @@ type Client struct {
 
 // Config holds configuration of the client.
 type Config struct {
-	// APIURL defaults to "https://api.kentik.com/api/v5"
+	// APIURL defaults to "https://api.kentik.com"
 	APIURL string
-	// SyntheticsHostPort defaults to "synthetics.api.kentik.com:443".
-	SyntheticsHostPort string
-	// CloudExportHostPort defaults to "cloudexport.api.kentik.com:443".
-	CloudExportHostPort string
-	AuthEmail           string
-	AuthToken           string
-	RetryCfg            RetryConfig
+
+	AuthEmail string
+	AuthToken string
+	RetryCfg  RetryConfig
 
 	// LogPayloads enables logging of request and response payloads to Cloud Export and Synthetics APIs.
 	// Note that this feature is currently broken.
@@ -78,9 +75,6 @@ type Config struct {
 	// A Timeout of zero means no timeout.
 	// Default: 100 seconds.
 	Timeout *time.Duration
-	// DisableTLS disables TLS for client connections.
-	// It has effect only on gRPC services: CloudExportAdmin, SyntheticsAdmin and SyntheticsData.
-	DisableTLS bool
 }
 
 type RetryConfig = httputil.RetryConfig
@@ -89,17 +83,17 @@ type RetryConfig = httputil.RetryConfig
 func NewClient(c Config) (*Client, error) {
 	c.FillDefaults()
 
-	syntheticsConnection, err := makeConnForGRPC(c, c.SyntheticsHostPort)
+	apiV5URL, err := makeAPIV5URL(c.APIURL)
 	if err != nil {
-		return nil, fmt.Errorf("grpc synthetics connection: %v", err)
-	}
-	cloudExportConnection, err := makeConnForGRPC(c, c.CloudExportHostPort)
-	if err != nil {
-		return nil, fmt.Errorf("grpc cloud export connection: %v", err)
+		return nil, fmt.Errorf("make API v5 URL: %v", err)
 	}
 
+	grpcConnection, err := makeConnForGRPC(c)
+	if err != nil {
+		return nil, fmt.Errorf("grpc connection: %v", err)
+	}
 	rc := api_connection.NewRestClient(api_connection.RestClientConfig{
-		APIURL:    c.APIURL,
+		APIURL:    apiV5URL,
 		AuthEmail: c.AuthEmail,
 		AuthToken: c.AuthToken,
 		RetryCfg:  c.RetryCfg,
@@ -107,7 +101,7 @@ func NewClient(c Config) (*Client, error) {
 	})
 	return &Client{
 		Alerting:           resources.NewAlertingAPI(rc),
-		CloudExports:       resources.NewCloudExportsAPI(cloudExportConnection),
+		CloudExports:       resources.NewCloudExportsAPI(grpcConnection),
 		CustomApplications: resources.NewCustomApplicationsAPI(rc),
 		CustomDimensions:   resources.NewCustomDimensionsAPI(rc),
 		DeviceLabels:       resources.NewDeviceLabelsAPI(rc),
@@ -120,8 +114,8 @@ func NewClient(c Config) (*Client, error) {
 		Tags:               resources.NewTagsAPI(rc),
 		Users:              resources.NewUsersAPI(rc),
 
-		SyntheticsAdmin: grpcsynthetics.NewSyntheticsAdminServiceClient(syntheticsConnection),
-		SyntheticsData:  grpcsynthetics.NewSyntheticsDataServiceClient(syntheticsConnection),
+		SyntheticsAdmin: grpcsynthetics.NewSyntheticsAdminServiceClient(grpcConnection),
+		SyntheticsData:  grpcsynthetics.NewSyntheticsDataServiceClient(grpcConnection),
 
 		config: c,
 	}, nil
@@ -132,14 +126,6 @@ func (c *Config) FillDefaults() {
 		c.APIURL = APIURLUS
 	}
 
-	if c.CloudExportHostPort == "" {
-		c.CloudExportHostPort = cloudExportHostPort
-	}
-
-	if c.SyntheticsHostPort == "" {
-		c.SyntheticsHostPort = syntheticsHostPort
-	}
-
 	if c.Timeout == nil {
 		c.Timeout = pointer.ToDuration(defaultTimeout)
 	}
@@ -147,10 +133,23 @@ func (c *Config) FillDefaults() {
 	c.RetryCfg.FillDefaults()
 }
 
-func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error) {
+func makeAPIV5URL(apiURL string) (string, error) {
+	u, err := url.Parse(apiURL)
+	if err != nil {
+		return "", err
+	}
+	u.Path += "/api/v5"
+	return u.String(), nil
+}
+
+func makeConnForGRPC(c Config) (grpc.ClientConnInterface, error) {
+	grpcHostPort, tlsEnabled, err := makeGRPCHostPort(c.APIURL)
+	if err != nil {
+		return nil, err
+	}
 	return grpc.Dial(
-		hostPort,
-		makeTLSOption(c),
+		grpcHostPort,
+		makeTLSOption(tlsEnabled),
 		grpc.WithUnaryInterceptor(
 			grpcmiddleware.ChainUnaryClient(
 				makeTimeoutInterceptor(c),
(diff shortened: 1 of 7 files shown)


kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
 	}
 
 	c.RetryCfg.FillDefaults()
 }
 
-func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error) {
+func parseKentikURL(apiURL *string) (string, bool) {
+	tls := false
+	u, err := url.Parse(*apiURL)
+	if err == nil {
```
mateuszmidor: It easier to read if you test for err != nil and return immediately

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
+		*apiURL += "/api/v5"
```
mateuszmidor, line 136: Should we only append/api/v5 if scheme is https?

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
-
 	if c.Timeout == nil {
 		c.Timeout = pointer.ToDuration(defaultTimeout)
 	}
 
 	c.RetryCfg.FillDefaults()
 }
 
-func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error) {
+func parseKentikURL(apiURL *string) (string, bool) {
```
mateuszmidor: Is apiURL an output parameter?

kentikapi/client.go
mateuszmidor: Does this mean we both update input apiURL and return output grpc?

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
+	if err == nil {
+		*apiURL += "/api/v5"
+		grpc := u.Host
+		if u.Scheme == "https" {
+			tls = true
+			if u.Port() == "" {
+				grpc += ":443"
+			}
+		}
+		return fmt.Sprintf("grpc." + grpc), tls
```
mateuszmidor: If I understand it correctly, this function transforms https URL in grpc. So there is implicit assumption that such mapping https->grpc always applies?

kentikapi/client.go
```diff
@@ -161,8 +160,8 @@ func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error
 	)
 }
 
-func makeTLSOption(c Config) grpc.DialOption {
-	if c.DisableTLS {
+func makeTLSOption(t bool) grpc.DialOption {
```
mateuszmidor: Please use something logner than "t", from function's signature it's not clear what it does :)

kentikapi/client.go
jdziurzynski: If input string is a valid url, function return url for grpc resources. If not, return input string directly.

kentikapi/client.go
jdziurzynski: Yes if input string is a valid url. This is a bit strange logic but needed for tests.  
Normally when user pass a valid url like https://api.kentik.com we want to have https://api.kentik.com/api/v5 and grpc.api.kentik.com:443 url parameters.
When we running test, input string will be local host with port like: 127.0.0.1:3869 This is not a valid url and should be api endpoint without modifications.

kentikapi/client.go
danielfurman: Agreed. I suggest "tlsEnabled"

kentikapi/client.go
```diff
@@ -78,24 +74,18 @@ type Config struct {
...
-	// It has effect only on gRPC services: CloudExportAdmin, SyntheticsAdmin and SyntheticsData.
-	DisableTLS bool
 }
 
 type RetryConfig = httputil.RetryConfig
 
 // NewClient creates a new Kentik API client.
 func NewClient(c Config) (*Client, error) {
 	c.FillDefaults()
+	grpcHostPort, tls := parseKentikURL(&c.APIURL)
```
danielfurman: Both arguments returned by parseKentikURL() are needed only in makeConnForGRPC() func. So, please move parseKentikURL() function call there.

kentikapi/client.go
```diff
@@ -119,9 +109,9 @@ func NewClient(c Config) (*Client, error) {
 		Tags:               resources.NewTagsAPI(rc),
 		Users:              resources.NewUsersAPI(rc),
 
-		CloudExportAdmin: grpccloudesxport.NewCloudExportAdminServiceClient(cloudExportConnection),
-		SyntheticsAdmin:  grpcsynthetics.NewSyntheticsAdminServiceClient(syntheticsConnection),
-		SyntheticsData:   grpcsynthetics.NewSyntheticsDataServiceClient(syntheticsConnection),
+		CloudExportAdmin: grpccloudesxport.NewCloudExportAdminServiceClient(grpcConnection),
```
danielfurman: This PR is not re-based onto latest upstream and conflicts with it. Please rebase it ASAP to save reviewers' time.

kentikapi/client.go
danielfurman:
```suggestion
func makeGRPCHostPort(apiURL string) (grpcHostPort string, tlsEnabled bool) {
```
1. This function does more than just parsing given URL.
2. No need for pointer to string
3. Named return parameters are great for making signature readable in such cases.

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
+			if u.Port() == "" {
+				grpc += ":443"
+			}
+		}
+		return fmt.Sprintf("grpc." + grpc), tls
+	}
+	return *apiURL, tls
+}
+
+func makeConnForGRPC(c Config, hostPort string, tls bool) (grpc.ClientConnInterface, error) {
```
danielfurman:
```suggestion
func makeConnForGRPC(c Config, hostPort string, tlsEnabled bool) (grpc.ClientConnInterface, error) {
```
Please unify naming to sth like `tlsEnabled`

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
+		grpc := u.Host
+		if u.Scheme == "https" {
+			tls = true
+			if u.Port() == "" {
+				grpc += ":443"
+			}
+		}
+		return fmt.Sprintf("grpc." + grpc), tls
+	}
+	return *apiURL, tls
```
danielfurman: I think we should return error when URL is not valid.

kentikapi/client.go
danielfurman: 1. Please split functions building API v5 and API v6 endpoints. We do not need unnecessary complexity.
2. Please modify tests to pass valid URLs, if needed. E.g. prepend `http://` scheme
3. Please add a logic that skip prepending `grpc` subdomain do the host when host is an IP.

kentikapi/client.go
```diff
@@ -132,25 +122,34 @@ func (c *Config) FillDefaults() {
...
+func parseKentikURL(apiURL *string) (string, bool) {
+	tls := false
+	u, err := url.Parse(*apiURL)
+	if err == nil {
+		*apiURL += "/api/v5"
+		grpc := u.Host
+		if u.Scheme == "https" {
+			tls = true
+			if u.Port() == "" {
+				grpc += ":443"
```
danielfurman: This should not be needed, because 443 is used by default: https://github.com/grpc/grpc/blob/master/doc/naming.md

kentikapi/client.go
danielfurman: Instead of building the URL string manually, please modify url object and then call String() on it.

danielfurman: Yes, there is such assumption: gRPC is served on `grpc` subdomain of Kentik API URL. It is documented in this chapter: https://docs.google.com/document/d/1b-huL-uHSVyI3eaao3-YOFHOZ115Sz9FOjlNOdHIULY/edit?pli=1#heading=h.w7r44tl9nlov
Also there are 2 resolved conversations in the design doc regarding URLs.

danielfurman (changes requested):

kentikapi/client.go
jdziurzynski: When trying Synthetics examples without specifying port:
rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp: address grpc.api.kentik.com: missing port in address"

kentikapi/client.go
```diff
@@ -132,25 +128,29 @@ func (c *Config) FillDefaults() {
...
-
 	if c.Timeout == nil {
 		c.Timeout = pointer.ToDuration(defaultTimeout)
 	}
 
 	c.RetryCfg.FillDefaults()
 }
 
-func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error) {
+func parseAPIURL(apiURL *string) error {
```
danielfurman:
```suggestion
func makeAPIV5URL(apiURL string) (string, error) {}
```
It is usually more readable to return results of operation as return function parameter instead of modifying input parameters. Also it is safer not to modify config.APIURL, which is then used to make gRPC URL, because it introduces coupling between those two pieces of logic. For example, changing hostname here would change hostname also for gRPC URL, which might not be desirable

Also, I see the naming issue repeated here. This function does more than parsing API URL. Hence its name should be different - I added suggestion.

kentikapi/client.go
```diff
@@ -89,15 +83,14 @@ type RetryConfig = httputil.RetryConfig
 func NewClient(c Config) (*Client, error) {
 	c.FillDefaults()
 
-	syntheticsConnection, err := makeConnForGRPC(c, c.SyntheticsHostPort)
-	if err != nil {
-		return nil, fmt.Errorf("grpc synthetics connection: %v", err)
+	if err := parseAPIURL(&c.APIURL); err != nil {
+		return nil, fmt.Errorf("invalid api url: %v", err)
```
danielfurman:
```suggestion
		return nil, fmt.Errorf("make API v5 URL: %v", err)
```
Error message should describe what the just-called function does. "invalid API URL: %v" message or "parse API URL: %v" would be appropriate next to url.Parse() call.
See also: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

kentikapi/client.go
```diff
@@ -161,8 +158,28 @@ func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error
 	)
 }
 
-func makeTLSOption(c Config) grpc.DialOption {
-	if c.DisableTLS {
+func makeGRPCHostPort(apiURL string) (grpcHostPort string, tlsEnabled bool, err error) {
+	tlsEnabled = false
+	u, err := url.Parse(apiURL)
+	if err != nil {
+		return apiURL, tlsEnabled, err
```
danielfurman:
```suggestion
		return "", false, err
```
It's better to return zero values of parameters in such case instead of arbitrary ones, like `apiURL`

kentikapi/client.go
```diff
@@ -161,8 +158,28 @@ func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error
...
+	}
+	grpc := u.Host
```
danielfurman:
```suggestion
	grpcHostPort = u.Host
```

kentikapi/client.go
```diff
@@ -132,25 +125,29 @@ func (c *Config) FillDefaults() {
...
 
 	c.RetryCfg.FillDefaults()
 }
 
-func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error) {
+func parseAPIURL(apiURL *string) error {
+	if _, err := url.Parse(*apiURL); err != nil {
+		return err
+	}
+	*apiURL += "/api/v5"
```
danielfurman: Please set u.Path to "/api/v5" and return u.String()

kentikapi/client.go
```diff
@@ -161,8 +158,28 @@ func makeConnForGRPC(c Config, hostPort string) (grpc.ClientConnInterface, error
...
+	tlsEnabled = false
+	u, err := url.Parse(apiURL)
+	if err != nil {
+		return apiURL, tlsEnabled, err
+	}
+	grpc := u.Host
+	if u.Scheme == "https" {
+		tlsEnabled = true
+		if u.Port() == "" {
+			grpc += ":443"
```
danielfurman: Why the code appends missing port only for https case? I think it should add :80 for http.

kentikapi/internal/testutil/http.go
```diff
@@ -34,22 +35,26 @@ func NewSpyHTTPHandler(t testing.TB, responseCode int, responseBody []byte) *Spy
...
-
-	err = r.Body.Close()
-	assert.NoError(h.t, err)
-
-	rw.Header().Set("Content-Type", "application/json")
-	rw.WriteHeader(h.responseCode)
-	_, err = rw.Write(h.responseBody)
-	assert.NoError(h.t, err)
+	if r.Method == "PRI" {
+		log.Printf("Omitting HTTP/2 connection preface (PRI)\n")
```
danielfurman:
```suggestion
		h.t.log.Printf("SpyHTTPHandler: omitting HTTP/2 connection preface (PRI)")
```
IMO it's better to use logger provided by testing package than importing new one. We use that style in all other tests.
Also, it's better to add context which part of testing code is logging. It gets confusing quickly when you don't remember the code exactly.
BTW: we found that those new lines are not needed

The same for MultipleResponseSpyHTTPHandler

kentikapi/internal/testutil/http.go
```diff
@@ -34,22 +35,26 @@ func NewSpyHTTPHandler(t testing.TB, responseCode int, responseBody []byte) *Spy
...
+	} else {
```
danielfurman: Please just call return in `if` block instead of adding `else` block.
The same for MultipleResponseSpyHTTPHandler

danielfurman (changes requested):
Couple comments

kentikapi/client.go
danielfurman: All 3 done

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go
danielfurman: Partially done; new comments below

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go
danielfurman: Ok.
> he following schemes are supported by the gRPC C-core implementation, but may not be supported in other languages: 
> ipv4:address[:port][,address[:port],...] -- IPv4 addresses

It seems that it is not the case for grpc-go.

kentikapi/client.go
danielfurman: Added comment for API v5 below.. For API v6 using url.String() is not desirable, indeed, because we need only host+port here.
This one is done.

danielfurman:
Branch is re-based incorrectly - there are now 2 "suggestion from review" commits. I suggest to rebase it with squash and name properly

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/client.go

kentikapi/internal/testutil/http.go
danielfurman: Done

kentikapi/internal/testutil/http.go
danielfurman: Done

danielfurman (approved):

mmac-m3a (approved):

PR #81, opened 2022-02-15
Provide consistent interface for CloudExport API

danielfurman:
Issue: KNTK-377

PR #88, opened 2022-03-18
Address Cloud Export API #81 review comments

danielfurman:
This PR addresses several review comments from PR #81 
Properties union is going to be added in separate PR.

Issue: KNTK-377


kentikapi/cloud_exports_integration_test.go
```diff
@@ -84,6 +84,28 @@ func TestClient_GetAllCloudExports(t *testing.T) {
 				},
 				InvalidCloudExportsCount: 1,
 			},
+		}, {
+			name: "2 exports received - one empty",
```
mmac-m3a, line 88: Is the API really returning empty entries? It should not ...

mmac-m3a (approved):

kentikapi/cloud_exports_integration_test.go
danielfurman: The generated gRPC client stub always initializes cloud export payload and SDK returns that data converted to model. We could add response validation to check required fields in response. This would catch possible bugs of the API server. Should I do that? It could be done along request validation which I planned to do as separate task.

PR #99, opened 2022-04-28
KentikError error type & set of functions to check error type

jdziurzynski:
KNTK-191

Errors returned by the api client are wrapped by new KentikError struct. This allows the user to check their type by using following functions:

func IsAuthError(err error) bool {}
func IsInvalidRequestError(err error) bool {}
func IsNotFoundError(err error) bool {}
func IsRateLimitExhaustedError(err error) bool {}
func IsTemporaryError(err error) bool {}
func IsTimeoutError(err error) bool {}

Full task description:
https://docs.google.com/document/d/1b-huL-uHSVyI3eaao3-YOFHOZ115Sz9FOjlNOdHIULY/edit#heading=h.6hi58cjyboxo


kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	Timeout            Code = 5
+)
+
+type KentikError struct {
+	Codes []Code
+	Msg   string
+}
+
+func (ke KentikError) Error() string {
+	return fmt.Sprintf("category: %v, message: %s \n", ke.Codes, ke.Msg)
```
mateuszmidor: not sure if newline should be included or left for user's decision?

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (ke KentikError) IsKentikError() *KentikError { return &ke }
```
mateuszmidor: Is* methods usually return bool
should this one be renamed?

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	return err
+}
+
+func KentikErrorFromHTTP(response *http.Response, err error) error {
+	if err == nil {
+		return nil
+	}
+	ke := KentikError{Msg: err.Error()}
+	if response == nil {
+		ke.Codes = []Code{Timeout}
```
mateuszmidor: might be easier to read, if method returned from here right away

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	"net/http"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Code int
+
+const (
+	AuthError          Code = 0
```
mateuszmidor: Can use iota for enum declaration

mateuszmidor (approved):

mmac-m3a (approved):
LGTM.

Needs to be rebased before merging.

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+func (ke KentikError) Timeout() bool {
+	for _, c := range ke.Codes {
+		if c == Timeout {
+			return true
+		}
+	}
+	return false
+}
+
+func IsAuthError(err error) bool {
```
danielfurman: Please put library public methods  (that are meant for users) in kentikapi/errors.go and rest of code preferably in kentikapi/internal/errors/errors.go. 
1. There is no need for separate package for several functions. Just like we didn't create separate package for credentials.go
2. That way the usage will be slightly less verbose: `kentikErrors.IsTimeoutError(err)` -> `kentikapi.IsTimeoutError(err)`
3. We put internal code in _internal_ subdirectory, so that unwanted code isn't used by library users. I don't see a reason for breaking that convention. 

Also, please add docstrings for library public methods.
Edit: apart from docstrings I think that we should write a section about inspecting errors in _README.md#usage_

danielfurman (changes requested):
Partial review

kentikapi/cloud_exports_integration_test.go
```diff
@@ -32,15 +33,15 @@ func TestClient_GetAllCloudExports(t *testing.T) {
 		response          listCEResponse
 		expectedResult    *models.GetAllCloudExportsResponse
 		expectedError     bool
-		expectedErrorCode *codes.Code
+		expectedErrorCode []kentikErrors.Code
```
danielfurman: Integration tests should test through public interfaces so that:
1. It is ensured that public interfaces work as expected and are usable
2. Change in implementation details do not affect tests

Public interface of error layer is set of `Is*Error()` functions. So I think we should check errors in those tests via those functions. It applies to all integration tests.
How to do that? We could add `tt.checkError(err)` function to test table or even better `errorPredicates []func(err) bool` that is a list of error predicates (error check functions) fulfilled by returned errror

kentikapi/errors/errors.go
danielfurman: I agree

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	Timeout            Code = 5
+)
+
+type KentikError struct {
+	Codes []Code
+	Msg   string
+}
+
+func (ke KentikError) Error() string {
+	return fmt.Sprintf("category: %v, message: %s", ke.Codes, ke.Msg)
```
danielfurman:
```suggestion
	return fmt.Sprintf("code: %v, message: %s", ke.Codes, ke.Msg)
```
Let's keep naming consistent

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	AuthError          Code = 0
+	InvalidRequest     Code = 1
+	NotFound           Code = 2
+	RateLimitExhausted Code = 3
+	Temporary          Code = 4
+	Timeout            Code = 5
+)
+
+type KentikError struct {
+	Codes []Code
```
danielfurman:
```suggestion
	code Code
```
I don't see a reason to hold multiple error codes. For example, RateLimitExhausted is temporary error, so there is no need to mark it also as Temporary.

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	Msg   string
```
danielfurman:
```suggestion
	msg   string
```

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+			ke.Codes = []Code{Timeout, Temporary}
+		case http.StatusTooManyRequests:
+			ke.Codes = []Code{RateLimitExhausted, Temporary}
+		case http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			ke.Codes = []Code{Temporary}
+		default:
+			ke.Codes = []Code{InvalidRequest}
+		}
+	} else {
+		ke.Codes = []Code{Timeout}
```
danielfurman: That case covers multiple possible errors: context.Canceled, context.DeadlineExceeded, errorString from retryablehttp...
Please set Timeout only on DeadlineExceeded error.
In other cases I suggest to set code Unknown. That way we explicitly inform user that we don't distinguish such error.

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		case http.StatusNotFound:
+			ke.Codes = []Code{NotFound}
+		case http.StatusRequestTimeout:
+			ke.Codes = []Code{Timeout, Temporary}
+		case http.StatusTooManyRequests:
+			ke.Codes = []Code{RateLimitExhausted, Temporary}
+		case http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			ke.Codes = []Code{Temporary}
+		default:
+			ke.Codes = []Code{InvalidRequest}
```
danielfurman:
```suggestion
			ke.Codes = []Code{Unknown}
```
I suggest to set code Unknown. That way we explicitly inform user that we don't distinguish such error.

kentikapi/internal/api_connection/rest_client.go
```diff
@@ -172,7 +185,8 @@ func (c *RestClient) Delete(ctx context.Context, path string) (responseBody json
 func errorFromResponseStatus(r *http.Response, responseBody string) error {
 	// TODO(dfurman): return more specific errors
```
danielfurman:
```suggestion
```

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		GetKentikError() *KentikError
+	}
+	if ok := errors.As(err, &tErr); ok {
+		ktError := tErr.GetKentikError()
+		return ktError.Codes, true
+	}
+	return []Code{}, false
+}
+
+func KentikErrorUpdateMsg(msg string, err error) error {
```
danielfurman: To add context information to the error, wrapping technique should be used. Please remove KentikErrorUpdateMsg() function. If I am missing something, please explain.

Example: `fmt.Errorf("cloud export with index %v: %w", i, err)`

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+const (
+	AuthError          Code = 0
+	InvalidRequest     Code = 1
+	NotFound           Code = 2
+	RateLimitExhausted Code = 3
+	Temporary          Code = 4
+	Timeout            Code = 5
+)
+
+type KentikError struct {
```
danielfurman: Please add New(code, msg) constructor for KentikError and use for creating new errors in the code.

kentikapi/internal/api_payloads/cloud_export.go
```diff
@@ -43,12 +45,18 @@ func cloudExportsFromPayload(exports []*cloudexportpb.CloudExport) ([]models.Clo
 // CloudExportFromPayload converts cloud export payload to model.
 func CloudExportFromPayload(ce *cloudexportpb.CloudExport) (*models.CloudExport, error) {
 	if ce == nil {
-		return nil, fmt.Errorf("cloud export response payload is nil")
+		return nil, kentikErrors.KentikError{
+			Codes: []kentikErrors.Code{kentikErrors.NotFound},
```
danielfurman:
```suggestion
			Codes: []kentikErrors.Code{kentikErrors.InvalidResponse},
```
This comes from invalid response format, so I suggest adding such status (only status; no check function for now).

kentikapi/internal/api_payloads/cloud_export.go
```diff
@@ -43,12 +45,18 @@ func cloudExportsFromPayload(exports []*cloudexportpb.CloudExport) ([]models.Clo
...
+			Msg:   "cloud export response payload is nil",
+		}
 	}
 
 	properties, err := propertiesFromPayload(ce)
 	if err != nil {
-		return nil, err
+		return nil, kentikErrors.KentikError{
+			Codes: []kentikErrors.Code{kentikErrors.NotFound},
```
danielfurman:
```suggestion
			Codes: []kentikErrors.Code{kentikErrors.InvalidResponse},
```

danielfurman: This is definitely not a NotFound error

kentikapi/internal/api_payloads/cloud_export.go
```diff
@@ -199,7 +207,10 @@ func cePayloadWithProperties(payload *cloudexportpb.CloudExport, ce *models.Clou
 	case "ibm":
 		payload.Properties = ibmPropertiesToPayload(ce)
 	default:
-		return nil, fmt.Errorf("invalid cloud provider: %v", ce.CloudProvider)
+		return nil, kentikErrors.KentikError{
+			Codes: []kentikErrors.Code{kentikErrors.NotFound},
```
danielfurman:
```suggestion
			Codes: []kentikErrors.Code{kentikErrors.InvalidRequest},
```

kentikapi/internal/utils/convert.go
```diff
@@ -34,7 +36,7 @@ func ConvertList(input interface{}, convertFunc ConvertFunc, output interface{})
 		results := vConvertFunc.Call(args)
 
 		if err, ok := results[1].Interface().(error); ok {
-			return err
+			return kentikErrors.KentikError{Codes: []kentikErrors.Code{kentikErrors.NotFound}, Msg: err.Error()}
```
danielfurman: I wouldn't touch this one. I doubt it is NotFound error from user's point of view.

danielfurman (changes requested):

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		}
+	}
+	return false
+}
+
+func IsAuthError(err error) bool {
+	var ktErr interface {
+		AuthError() bool
+	}
+	if ok := errors.As(err, &ktErr); ok {
```
danielfurman: As we're using only 1 custom error type - KentikError, we could just use it here instead of custom interface. That way we can retrieve error code with KenitkError.GetCode() and use it here.

All `Is*Error()` functions implementation can be simplified that way.
We can also remove methods such as `func (ke KentikError) AuthError() bool`. Those are not needed, because users are expected to call `Is*Error()` functions. (If we wouldn't provide check functions, then such methods would be needed.)

danielfurman (changes requested):

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+type KentikError struct {
+	Codes []Code
+	Msg   string
+}
+
+func (ke KentikError) Error() string {
+	return fmt.Sprintf("category: %v, message: %s", ke.Codes, ke.Msg)
+}
+
+func (ke KentikError) GetKentikError() *KentikError { return &ke }
```
danielfurman: IMO this method does not add any value. Such methods are useful, when we have multiple error types being in the same category, e.g. Temporary(), but we are using error type, so this is not the case. Please remove that method

In our internal errutil code we can just look for KentikError type in error chain directly.

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+
+func GetCodes(err error) ([]Code, bool) {
```
danielfurman: If we provide Code() Code getter method for KentikError, this function is not needed. Please remove it.

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	return err
+}
+
+func KentikErrorFromHTTP(response *http.Response, err error) error {
+	if err == nil {
+		return nil
+	}
+	ke := KentikError{Msg: err.Error()}
+	if response != nil {
+		switch response.StatusCode {
```
danielfurman: Please add those cases:
- HTTP 400 Bad Request -> InvalidRequest
- HTTP 409 Conflict -> IsInvalidRequestError

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		case codes.NotFound:
+			ke.Codes = []Code{NotFound}
+		case codes.PermissionDenied, codes.Unauthenticated:
+			ke.Codes = []Code{AuthError}
+		case codes.ResourceExhausted:
+			ke.Codes = []Code{RateLimitExhausted, Temporary}
+		case codes.Unavailable:
+			ke.Codes = []Code{Temporary}
+		default:
+			ke.Codes = []Code{InvalidRequest}
```
danielfurman:
```suggestion
			ke.Codes = []Code{Unknown}
```

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return ke
+}
+
+func KentikErrorFromGRPC(err error) error {
+	if err == nil {
+		return nil
+	}
+	ke := KentikError{Msg: err.Error()}
+	if e, ok := status.FromError(err); ok {
```
danielfurman: Please add following cases:
- InvalidArgument / AlreadyExists / FailedPrecondition / OutOfRange -> InvalidRequest

danielfurman (changes requested):

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		case codes.DeadlineExceeded:
+			ke.Codes = []Code{Timeout, Temporary}
+		case codes.NotFound:
+			ke.Codes = []Code{NotFound}
+		case codes.PermissionDenied, codes.Unauthenticated:
+			ke.Codes = []Code{AuthError}
+		case codes.ResourceExhausted:
+			ke.Codes = []Code{RateLimitExhausted, Temporary}
+		case codes.Unavailable:
+			ke.Codes = []Code{Temporary}
```
danielfurman:
```suggestion
			ke.Codes = []Code{Unavailable}
```
I know this is temporary error, but I'd rather keep root cause in error code, so that this info is not lost

kentikapi/errors/errors.go
```diff
@@ -0,0 +1,231 @@
...
+		case http.StatusUnauthorized, http.StatusForbidden:
+			ke.Codes = []Code{AuthError}
+		case http.StatusNotFound:
+			ke.Codes = []Code{NotFound}
+		case http.StatusRequestTimeout:
+			ke.Codes = []Code{Timeout, Temporary}
+		case http.StatusTooManyRequests:
+			ke.Codes = []Code{RateLimitExhausted, Temporary}
+		case http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			ke.Codes = []Code{Temporary}
```
danielfurman:
```suggestion
			ke.Codes = []Code{Unavailable}
```
I know this is temporary error, but I'd rather keep root cause in error code so that this info is not lost

danielfurman (changes requested):

kentikapi/client_integration_test.go
```diff
@@ -2,7 +2,6 @@ package kentikapi_test
```
danielfurman, line 2: Please verify returned errors where applicable via public `Is*Error()` checks in:
- TestClient_GetUserWithRetries
- TestClient_GetAgentWithRetries
- all tests in synth_agents_integration_test.go (update the branch before)
- users_integration_test.go, tags_integration_test.go (unless you have reason why not)

Ideally extend TestClient tests to test several HTTP/gRPC error conversions.

BTW, we need error translations in just-merged synthetics agents code.

danielfurman (changes requested):
Review finished

kentikapi/errors/errors.go
jdziurzynski: My knowledge about this task is based on https://docs.google.com/document/d/1b-huL-uHSVyI3eaao3-YOFHOZ115Sz9FOjlNOdHIULY/edit#heading=h.6hi58cjyboxo
To be right with the error table from this document I decide to keep error codes in slice.

kentikapi/errors/errors.go
jdziurzynski: There is no such an error code as DeadlineExceeded in net/http package.
This switch statement is based on error table from https://docs.google.com/document/d/1b-huL-uHSVyI3eaao3-YOFHOZ115Sz9FOjlNOdHIULY/edit#heading=h.6hi58cjyboxo. 
If You think that the table need some corrections please update it and I will update errors in code. It will be easier for me.

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
+package kentikapi
+
+import (
+	"errors"
+
+	kentikError "github.com/kentik/community_sdk_golang/kentikapi/internal/errors"
```
danielfurman:
```suggestion
	kentikErrors "github.com/kentik/community_sdk_golang/kentikapi/internal/errors"
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+)
+
+// IsAuthError checks if passed error is Authorization error, if yes return true.
+func IsAuthError(err error) bool {
+	var ktErr interface {
```
danielfurman:
```suggestion
	var ke kentikErrors.KentikError
```
There is no need to look for any error with Code() method. We can simplify it to look for specific KentikError.
In other places too.

kentikapi/errors.go
danielfurman: As we're on this. All other import aliases in the repo use all lowercase style, e.g. syntheticspb. What do you think about renaming it to match that style, e.g. `kentikerrors` or `kerrors`?

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+	NotFound           Code = 3
+	RateLimitExhausted Code = 4
+	Temporary          Code = 5
+	Timeout            Code = 6
+	Unavailable        Code = 7
+	Unknown            Code = 8
+	InvalidResponse    Code = 9
+)
+
+type KentikError struct {
```
danielfurman:
```suggestion
type StatusError struct {
```
I know that I didn't pointed it out earlier, but I had more important comments to make. I think that adding _Kentik_ prefix to the type located in _kentikapi_ module is not useful. However, StatusError tells that the error contains status code, which describes the purpose of the error type.

Those renames shouldn't be hard using an IDE.

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+const (
+	AuthError          Code = 1
+	InvalidRequest     Code = 2
+	NotFound           Code = 3
+	RateLimitExhausted Code = 4
+	Temporary          Code = 5
+	Timeout            Code = 6
+	Unavailable        Code = 7
+	Unknown            Code = 8
+	InvalidResponse    Code = 9
```
danielfurman: Nit: Please put it next to InvalidRequest

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+	"net/http"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type Code int
+
+const (
+	AuthError          Code = 1
```
danielfurman:
```suggestion
	AuthError          Code = iota
```
Please use iota and drop values below for simplicity, as it was pointed out earlier. See: https://go.dev/doc/effective_go#constants

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+		msg:  msg,
+	}
+}
+
+func (ke KentikError) Error() string {
+	return fmt.Sprintf("code: %v, message: %s", ke.code, ke.msg)
+}
+
+func (ke KentikError) Code() Code {
+	if ke.code == 0 {
```
danielfurman:
```suggestion
```
Currently, if error is wrongly initialized, the code returned is not Unknown. And I think it should be Unknown. To guard against that please put Unknown as the first code in enum const() block and assign iota to it (which is 0).

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+		case http.StatusTooManyRequests:
+			ke.code = RateLimitExhausted
+		case http.StatusServiceUnavailable, http.StatusGatewayTimeout:
+			ke.code = Unavailable
+		}
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		ke.code = Timeout
+	}
+	if ke.code == 0 {
```
danielfurman:
```suggestion
```
This _if_ can be removed, when ke initialization sets code to Unknown.

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,104 @@
...
+		case codes.ResourceExhausted:
+			ke.code = RateLimitExhausted
+		case codes.Unavailable:
+			ke.code = Unavailable
+		case codes.InvalidArgument, codes.AlreadyExists, codes.FailedPrecondition, codes.OutOfRange:
+			ke.code = InvalidRequest
+		default:
+			ke.code = Unknown
+		}
+	} else {
```
danielfurman: This _else_ can be removed when ke initialization sets code to Unknown.

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
+package kentikapi
+
+import (
+	"errors"
+
+	kentikError "github.com/kentik/community_sdk_golang/kentikapi/internal/errors"
+)
+
+// IsAuthError checks if passed error is Authorization error, if yes return true.
```
danielfurman:
```suggestion
// IsAuthError returns true if passed error is an authentication/authorization error.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+	var ktErr interface {
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.AuthError
+	}
+	return false
+}
+
+// IsInvalidRequestError checks if passed error is invalid request error, if yes return true.
```
danielfurman:
```suggestion
// IsInvalidRequestError returns true if passed error is due to invalid request.
// It might be returned if user passed invalid data or there is a resource conflict.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+	var ktErr interface {
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.InvalidRequest
+	}
+	return false
+}
+
+// IsNotFoundError checks if passed error is not found error, if yes return true.
```
danielfurman:
```suggestion
// IsNotFoundError returns true if passed error indicates that resource was not found.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+	var ktErr interface {
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.NotFound
+	}
+	return false
+}
+
+// IsRateLimitExhaustedError checks if passed error is rate limit exhausted error, if yes return true.
```
danielfurman:
```suggestion
// IsRateLimitExhaustedError returns true if passed error indicates that the rate limit for a resource has been exhausted.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+	var ktErr interface {
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.RateLimitExhausted
+	}
+	return false
+}
+
+// IsTemporaryError checks if passed error is temporary error, if yes return true.
```
danielfurman:
```suggestion
// IsTemporaryError returns true if passed error is temporary, e.g. it is due to a timeout, rate limit exhaustion or service unavailability.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+		if ktErr.Code() == kentikError.Temporary ||
+			ktErr.Code() == kentikError.Timeout ||
+			ktErr.Code() == kentikError.RateLimitExhausted {
+			return true
+		}
+	}
+	return false
+}
+
+// IsTimeoutError checks if passed error is timeout error, if yes return true.
```
danielfurman:
```suggestion
// IsTimeoutError returns true if passed error is a timeout error.
```

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.Unavailable
+	}
+	return false
+}
+
+// IsUnknownError checks if passed error is unknown error, if yes return true.
+func IsUnknownError(err error) bool {
```
danielfurman:
```suggestion
```
What's the purpose of the function? I don't see a use case for that.

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		return ktErr.Code() == kentikError.Timeout
+	}
+	return false
+}
+
+// IsUnavailableError checks if passed error is unavailable error, if yes return true.
+func IsUnavailableError(err error) bool {
```
danielfurman:
```suggestion
```
Please remove that. I don't think that this function is necessary and I'd like to keep the API as small as possible.  It can always be added, if needed.
Note that it was not included in the design.

kentikapi/errors.go
```diff
@@ -0,0 +1,99 @@
...
+// IsTemporaryError checks if passed error is temporary error, if yes return true.
+func IsTemporaryError(err error) bool {
+	var ktErr interface {
+		Code() kentikError.Code
+	}
+	if ok := errors.As(err, &ktErr); ok {
+		if ktErr.Code() == kentikError.Temporary ||
+			ktErr.Code() == kentikError.Timeout ||
+			ktErr.Code() == kentikError.RateLimitExhausted {
+			return true
```
danielfurman, line 55: Please return true also for `Unavailable` code (just like in the design).

danielfurman (changes requested):

kentikapi/errors/errors.go
danielfurman: Done all except adding README.md#usage

kentikapi/cloud_exports_integration_test.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/internal/api_connection/rest_client.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/internal/api_payloads/cloud_export.go
danielfurman: Done

kentikapi/internal/api_payloads/cloud_export.go
danielfurman: Done

kentikapi/internal/api_payloads/cloud_export.go
danielfurman: Done

kentikapi/internal/utils/convert.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: First part of the comment is still to do; created separate comment for that: https://github.com/kentik/community_sdk_golang/pull/99#discussion_r877928236

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/client_integration_test.go
```diff
@@ -149,9 +148,9 @@ func TestClient_GetUserWithRetries(t *testing.T) {
 			}
```
danielfurman, line 176: `default timeout is longer than 30 ms` and `timeout is longer than the wait for response with retries` could (and IMO should) verify that error is bad request.

Also `"timeout during first request"` could verify both isTimeoutError() and isTemporaryError() predicates.

kentikapi/client_integration_test.go
```diff
@@ -201,9 +200,9 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
 				newErrorGRPCGetAgentResponse(codes.Unavailable),
 				newErrorGRPCGetAgentResponse(codes.Unavailable),
 			},
-			expectedResult:    nil,
-			expectedError:     true,
-			expectedErrorCode: codePtr(codes.Unavailable),
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsUnavailableError},
```
danielfurman:
```suggestion
			errorPredicates: []func(error) bool{kentikapi.IsTemporaryError},
```

kentikapi/client_integration_test.go
```diff
@@ -213,26 +212,26 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
 			options: []kentikapi.ClientOption{kentikapi.WithRetryMaxAttempts(0)},
 			responses: []gRPCGetAgentResponse{
 				newErrorGRPCGetAgentResponse(codes.Unavailable),
 			},
-			expectedResult:    nil,
-			expectedError:     true,
-			expectedErrorCode: codePtr(codes.Unavailable),
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsUnavailableError},
```
danielfurman:
```suggestion
			errorPredicates: []func(error) bool{kentikapi.IsTemporaryError},
```

kentikapi/client_integration_test.go
```diff
@@ -241,17 +240,17 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
 				newErrorGRPCGetAgentResponse(codes.Unavailable),
 				newErrorGRPCGetAgentResponse(codes.Unavailable),
 			},
-			expectedResult:    nil,
-			expectedError:     true,
-			expectedErrorCode: codePtr(codes.Unavailable),
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsUnavailableError},
```
danielfurman:
```suggestion
			errorPredicates: []func(error) bool{kentikapi.IsTemporaryError},
```

kentikapi/client_integration_test.go
```diff
@@ -241,17 +240,17 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
-			expectedResult:    nil,
-			expectedError:     true,
-			expectedErrorCode: codePtr(codes.DeadlineExceeded),
+			name:            "timeout during first request",
+			options:         []kentikapi.ClientOption{kentikapi.WithTimeout(5 * time.Millisecond)},
+			responses:       []gRPCGetAgentResponse{},
+			handlingDelay:   1 * time.Second,
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsTimeoutError},
```
danielfurman:
```suggestion
			errorPredicates: []func(error) bool{kentikapi.IsTimeoutError, kentikapi.IsTemporaryError},
```

kentikapi/client_integration_test.go
danielfurman: Because I suggested to remove IsUnavaliableError() function.

kentikapi/client_integration_test.go
danielfurman: To verify that IsTemporaryError() function works correctly.

danielfurman (changes requested):

kentikapi/client_integration_test.go
danielfurman: I ran coverage test and here are my suggestions:

Please add subtests to both TestClient_GetUserWithRetries and TestClient_GetAgentWithRetries that cover IsAuthError() and IsRateLimitExhaustedError() cases. (We could put them in separate test in client_integration_test.go file, but that would be more work).
If we add subtests mentioned above, we have all check function tested and I am ok with skipping modifications to users_integration_test.go and tags_integration_test.go.

danielfurman (changes requested):
Final comment

kentikapi/client_integration_test.go
```diff
@@ -116,6 +119,29 @@ func TestClient_GetUserWithRetries(t *testing.T) {
 			timeout:             5 * time.Millisecond,
 			expectedError:       true,
 			expectedTimeout:     true,
+			errorPredicates:     []func(error) bool{kentikapi.IsTimeoutError, kentikapi.IsTemporaryError},
+		}, {
+			name: "authorization error",
+			responses: []testutil.HTTPResponse{
+				testutil.NewErrorHTTPResponse(http.StatusUnauthorized),
+			},
+			serverHandlingDelay: 10 * time.Millisecond,
```
danielfurman:
```suggestion
```
Why? Let's keep the sub-test lean.

kentikapi/client_integration_test.go
```diff
@@ -116,6 +119,29 @@ func TestClient_GetUserWithRetries(t *testing.T) {
...
+		}, {
+			name: "too many requests error",
+			responses: []testutil.HTTPResponse{
+				testutil.NewErrorHTTPResponse(http.StatusTooManyRequests),
+				testutil.NewErrorHTTPResponse(http.StatusTooManyRequests),
+				testutil.NewErrorHTTPResponse(http.StatusTooManyRequests),
+				testutil.NewErrorHTTPResponse(http.StatusTooManyRequests),
+				testutil.NewErrorHTTPResponse(http.StatusTooManyRequests),
+			},
+			serverHandlingDelay: 10 * time.Millisecond,
```
danielfurman:
```suggestion
```
Ditto

kentikapi/client_integration_test.go
```diff
@@ -242,15 +269,35 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
+			name: "authorization error",
+			responses: []gRPCGetAgentResponse{
+				newErrorGRPCGetAgentResponse(codes.Unauthenticated),
+			},
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsAuthError},
+		}, {
+			name:    "too many requests error",
+			options: []kentikapi.ClientOption{kentikapi.WithRetryMaxAttempts(3)},
```
danielfurman:
```suggestion
```
KISS

kentikapi/client_integration_test.go
```diff
@@ -242,15 +269,35 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
+		}, {
+			name: "authorization error",
+			responses: []gRPCGetAgentResponse{
+				newErrorGRPCGetAgentResponse(codes.Unauthenticated),
+			},
+			expectedResult:  nil,
+			expectedError:   true,
+			errorPredicates: []func(error) bool{kentikapi.IsAuthError},
+		}, {
+			name:    "too many requests error",
```
danielfurman:
```suggestion
			name:    "ResourceExhausted error",
```

kentikapi/client_integration_test.go
```diff
@@ -242,15 +269,35 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
 			name:            "timeout during first request",
 			options:         []kentikapi.ClientOption{kentikapi.WithTimeout(5 * time.Millisecond)},
 			responses:       []gRPCGetAgentResponse{},
 			handlingDelay:   1 * time.Second,
 			expectedResult:  nil,
 			expectedError:   true,
-			errorPredicates: []func(error) bool{kentikapi.IsTimeoutError},
+			errorPredicates: []func(error) bool{kentikapi.IsTimeoutError, kentikapi.IsTemporaryError},
+		}, {
+			name: "authorization error",
```
danielfurman:
```suggestion
			name: "Unauthenticated error",
```
authorization is not equal authentication (but we use AuthError for both)

kentikapi/cloud_exports_integration_test.go
```diff
@@ -319,14 +319,9 @@ func TestClient_CreateCloudExport(t *testing.T) {
...
 			request:         nil,
-			expectedRequest: &cloudexportpb.CreateCloudExportRequest{Export: nil},
-			response: createCEResponse{
-				data: &cloudexportpb.CreateCloudExportResponse{},
-				err:  status.Errorf(codes.InvalidArgument, codes.InvalidArgument.String()),
-			},
+			expectedRequest: nil,
 			expectedResult:  nil,
 			expectedError:   true,
-			errorPredicates: []func(error) bool{kentikapi.IsInvalidRequestError},
```
danielfurman: Why this one has been removed? No request data is indeed a invalid request from user's point of view

README.md
```diff
@@ -37,6 +37,10 @@ export KTAPI_AUTH_TOKEN=<Kentik API authentication token>
 go test -tags examples -count 1 -parallel 1 -v -run Users github.com/kentik/community_sdk_golang/examples
 ```
 
+### Errors
+
+Library provide a set of [functions](./kentikapi/errors.go) to check error type returned by the client.
```
danielfurman:
```suggestion
Library provides a set of [functions](./kentikapi/errors.go) to check error type returned by the client.
```

kentikapi/cloud_exports_integration_test.go
```diff
@@ -547,13 +543,9 @@ func TestClient_UpdateCloudExport(t *testing.T) {
...
 			name:            "nil request",
 			request:         nil,
-			expectedRequest: &cloudexportpb.UpdateCloudExportRequest{Export: nil},
-			response: updateCEResponse{
-				err: status.Errorf(codes.InvalidArgument, codes.InvalidArgument.String()),
-			},
+			expectedRequest: nil,
 			expectedResult:  nil,
 			expectedError:   true,
-			errorPredicates: []func(error) bool{kentikapi.IsInvalidRequestError},
```
danielfurman: Ditto (Why this one has been removed? No request data is indeed a invalid request from user's point of view)

kentikapi/internal/errors/errors.go
```diff
@@ -13,45 +13,42 @@ import (
...
-	InvalidResponse    Code = 9
+	Unknown Code = iota
+	AuthError
+	InvalidRequest
+	Timeout
+	Unavailable
+	InvalidResponse
+	NotFound
+	RateLimitExhausted
+	Temporary
```
danielfurman:
```suggestion
```
We don't need that status code It can always be added if we need it

kentikapi/internal/errors/errors.go
danielfurman: Closer but not quite done

kentikapi/errors/errors.go
danielfurman: Done

kentikapi/client_integration_test.go
danielfurman: Done

kentikapi/errors.go
danielfurman: I guess you agree with me :) Done

kentikapi/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done (removed)

kentikapi/errors.go
danielfurman: Done

kentikapi/errors.go
danielfurman: Done

kentikapi/client_integration_test.go
danielfurman: Done

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go
```diff
@@ -279,25 +277,24 @@ func TestClient_GetAgentWithRetries(t *testing.T) {
...
 			options: []kentikapi.ClientOption{kentikapi.WithRetryMaxAttempts(3)},
 			responses: []gRPCGetAgentResponse{
 				newErrorGRPCGetAgentResponse(codes.ResourceExhausted),
 				newErrorGRPCGetAgentResponse(codes.ResourceExhausted),
 				newErrorGRPCGetAgentResponse(codes.ResourceExhausted),
 				newErrorGRPCGetAgentResponse(codes.ResourceExhausted),
 			},
-			expectedResult:  nil,
-			expectedError:   true,
-			errorPredicates: []func(error) bool{kentikapi.IsRateLimitExhaustedError, kentikapi.IsTemporaryError},
```
danielfurman: The purpose of this test was to check if returned error satisfies IsRateLimitExhaustedError and IsTemporaryError predicates. And you have just removed that... It is better to fix the test instead of practically disabling it.
Tip: you need to add 1 more `newErrorGRPCGetAgentResponse(codes.ResourceExhausted),` response for test to be correct, because default RetryMaxAttempts is 4

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go
danielfurman: Still to do. Note that you need to adjust number of request for the test to work.
Another option is to disable retries in this subtest. But setting it to arbitrary value, like 3 is confusing to me (hence I made that comment)

kentikapi/client_integration_test.go

kentikapi/client_integration_test.go

kentikapi/cloud_exports_integration_test.go
danielfurman: Done

README.md
danielfurman: Done

kentikapi/cloud_exports_integration_test.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/internal/errors/errors.go
danielfurman: Still to do

kentikapi/client_integration_test.go
danielfurman: Still to do for analogous sub-test: "ResourceExhausted error"

kentikapi/client_integration_test.go
danielfurman: Still to do

danielfurman (changes requested):
Please address those issues:
- https://github.com/kentik/community_sdk_golang/pull/99#discussion_r883777078
- https://github.com/kentik/community_sdk_golang/pull/99#discussion_r878003685

kentikapi/internal/errors/errors.go
danielfurman: Done

kentikapi/client_integration_test.go

danielfurman (approved):

PR #102, opened 2022-06-01
Introduce synthetics and cloud domains

danielfurman:
Introduce synthetics and cloud domains                                                                                                                                                                                                    
                                                                                                                                                                                                                                              
This change is a refectoring of a package structure for synthetics and                                                                                                                                                                    
cloud exports code.                                                                                                                                                                                                                       
                                                                                                                                                                                                                                          
As the codebase grows larger, there are drawbacks of keeping packages split                                                                                                                                                               
to layers for all domains (models, resources, api_payloads, etc.).                                                                                                                                                                        
For example, I found that there is a need to prefix model types with                                                                                                                                                                      
the name of their domain for namespacing purposes - to avoid name conflicts.                                                                                                                                                              
Extracting models to domain packages (synthetics, cloud) fixes that problem and changes the usage                                                                                                                                         
of exported types, e.g. models.SyntheticsAgent changes to synthetics.Agent.                                                                                                                                                               
                                                                                                                                                                                                                                          
The same applies to internal packages. It is easier to work with multiple                                                                                                                                                                 
domain-specific packages instead of layer-based mega packages                                                                                                                                                                             
(resources, api_payloads).                                                                                                                                                                                                                
                                                                                                                                                                                                                                          
I propose to keep this domain-base style for all new APIs.                                                                                                                                                                                
 
Issue: KNTK-399
Based-on: #99 (to avoid conflicts)


kentikapi/client.go
```diff
@@ -52,12 +54,11 @@ type Client struct {
...
 	SavedFilters       *resources.SavedFiltersAPI
 	Sites              *resources.SitesAPI
-	Synthetics         *resources.SyntheticsAPI
+	Synthetics         *synthetics.API
 	Tags               *resources.TagsAPI
 	Users              *resources.UsersAPI
 
-	// SyntheticsAdmin and SyntheticsData are gRPC clients
-	// for Kentik API Cloud Export and Synthetics services.
+	// SyntheticsAdmin and SyntheticsData are legacy gRPC clients for Kentik API Synthetics services.
```
mmac-m3a, line 61: What is "legacy" about `syntheticspb.SyntheticsAdminServiceClient` and `syntheticspb.SyntheticsDataServiceClient`? Aren't those gRPC clients generated from the protobuf schema? If so, those are definitely not "legacy".

kentikapi/cloud/export.go
```diff
@@ -184,78 +186,78 @@ type CloudExportStatus struct {
...
+// Provider is the name of cloud provider.
+type Provider string
 
 const (
-	CloudProviderAlibaba CloudProvider = "alibaba"
-	CloudProviderAWS     CloudProvider = "aws"
-	CloudProviderAzure   CloudProvider = "azure"
-	CloudProviderGCE     CloudProvider = "gce" // gcp value in Agents API
-	CloudProviderIBM     CloudProvider = "ibm"
+	ProviderAlibaba Provider = "alibaba"
```
mmac-m3a, line 258: (unrelated to the change in this review) I think we can drop Alibaba. We do not support it yet and it is unclear if/when we will.

kentikapi/internal/cloud/cloud.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/kentik/community_sdk_golang/kentikapi/models"
+	"google.golang.org/grpc"
+)
+
+// API aggregates Cloud Exports API methods.
+type API struct {
+	client cloudexportpb.CloudExportAdminServiceClient
+}
+
+// NewAPI creates new API.
```
mmac-m3a, line 18: Is this useless (but harmless) comment forced by a linter?

kentikapi/internal/errors/errors.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func New(code Code, msg string) StatusError {
+	return StatusError{
+		code: code,
+		msg:  msg,
+	}
+}
+
+func (ke StatusError) Error() string {
```
mmac-m3a: What does `ke` stand for? Kentik Error, perhaps? I suggest to use `api_error` or just `e` (if linters are OK with that)

mmac-m3a (approved):
I like the new structure!

mmac-m3a:
Please, resolve conflicts and I'll merge the PR.

kentikapi/client.go
danielfurman: It might not be a good word, but it is better to use new client.Synthetics API than those generated clients directly. Those clients (and comment) will be removed soon.

kentikapi/cloud/export.go
danielfurman: I'll remove it in next PR

kentikapi/internal/cloud/cloud.go
danielfurman: It is not enforced by a linter, but it is a convention in Go to write comment for every exported symbol (https://go.dev/doc/effective_go#commentary). We reduced this convention to "every exported symbol used by SDK users". Those comments are generated into pkg.go.dev documentation and might not be so useless when there is no code visible (only function signature). It looks like this: https://pkg.go.dev/github.com/kentik/community_sdk_golang@v0.2.0/kentikapi#NewClient

This particular function is actually not going to be used by SDK users so is indeed useless

kentikapi/internal/errors/errors.go
danielfurman: The type was initially KentikError (which was a bad name), so `ke` made sense. It was not adjusted to the type's new name. I'll fix that in next PR

PR #103, opened 2022-06-01
Provide consistent interface for Synthetics Tests API

danielfurman:
GetTestResults() and GetTestTrace() methods will be implemented in next PR.
    
Based-on: https://github.com/kentik/community_sdk_golang/pull/104
Issue: KNTK-399

### DIFF
--- a/examples/synthetics_example_test.go
+++ b/examples/synthetics_example_test.go
@@ -544,7 +544,7 @@ func newMinimalDNSTest(agentIDs []models.ID) *synthetics.Test {
 			Timeout:    time.Minute,
 			RecordType: synthetics.DNSRecordAAAA,
 			Servers:    []net.IP{net.ParseIP("192.0.2.213"), net.ParseIP("2001:db8:dead:beef:dead:beef:dead:beef")},
-			Port:       443,
+			Port:       53,
 		},
 	})
 }
@@ -560,7 +560,7 @@ func newMinimalDNSGridTest(agentIDs []models.ID) *synthetics.Test {
 			Timeout:    time.Minute,
 			RecordType: synthetics.DNSRecordAAAA,
 			Servers:    []net.IP{net.ParseIP("192.0.2.213"), net.ParseIP("2001:db8:dead:beef:dead:beef:dead:beef")},
-			Port:       443,
+			Port:       53,
 		},
 	})
 }
--- a/examples/synthetics_example_test.go
+++ b/examples/synthetics_example_test.go
@@ -223,6 +223,7 @@ func newHostnameTest(ctx context.Context, client *kentikapi.Client) (*synthetics
 				Timeout:  10 * time.Second,
 				Count:    10,
 				Protocol: synthetics.PingProtocolTCP,
+				Port:     65535,
 			},
 			Traceroute: synthetics.TracerouteSettingsRequiredFields{
 				Timeout:  59999 * time.Millisecond,
@@ -350,6 +351,7 @@ func newMinimalIPTest(agentIDs []models.ID) *synthetics.Test {
 				Timeout:  10 * time.Second,
 				Count:    10,
 				Protocol: synthetics.PingProtocolTCP,
+				Port:     65535,
 			},
 			Traceroute: synthetics.TracerouteSettingsRequiredFields{
 				Timeout:  59999 * time.Millisecond,
@@ -376,6 +378,7 @@ func newMinimalNetworkGridTest(agentIDs []models.ID) *synthetics.Test {
 				Timeout:  10 * time.Second,
 				Count:    10,
 				Protocol: synthetics.PingProtocolTCP,
+				Port:     65535,
 			},
 			Traceroute: synthetics.TracerouteSettingsRequiredFields{
 				Timeout:  59999 * time.Millisecond,
@@ -402,6 +405,7 @@ func newMinimalHostnameTest(agentIDs []models.ID) *synthetics.Test {
 				Timeout:  10 * time.Second,
 				Count:    10,
 				Protocol: synthetics.PingProtocolTCP,
+				Port:     65535,
 			},
 			Traceroute: synthetics.TracerouteSettingsRequiredFields{
 				Timeout:  59999 * time.Millisecond,
--- a/examples/synthetics_example_test.go
+++ b/examples/synthetics_example_test.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
+	"math/rand"
 	"net"
 	"net/http"
 	"net/url"
@@ -16,6 +18,7 @@ import (
 
 	syntheticspb "github.com/kentik/api-schema-public/gen/go/kentik/synthetics/v202202"
 	"github.com/kentik/community_sdk_golang/kentikapi"
+	"github.com/kentik/community_sdk_golang/kentikapi/models"
 	"github.com/kentik/community_sdk_golang/kentikapi/synthetics"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -67,9 +70,10 @@ func demonstrateSyntheticsAgentAPI() error {
 	fmt.Println("Number of agents:", len(getAllResp.Agents))
 	fmt.Println("Number of invalid agents:", getAllResp.InvalidAgentsCount)
 
-	agentID, err := pickPrivateAgentID(ctx)
+	// Pick a private agent, so it is possible to modify it
+	agentID, err := pickPrivateAgentID(getAllResp.Agents)
 	if err != nil {
-		return fmt.Errorf("pick agent ID: %w", err)
+		return fmt.Errorf("pick private agent ID: %w", err)
 	}
 
 	fmt.Println("### Getting synthetics agent with ID", agentID)
@@ -126,27 +130,6 @@ func demonstrateSyntheticsAgentAPI() error {
 	return nil
 }
 
-func pickPrivateAgentID(ctx context.Context) (string, error) {
-	client, err := NewClient()
-	if err != nil {
-		return "", err
-	}
-
-	getAllResp, err := client.SyntheticsAdmin.ListAgents(ctx, &syntheticspb.ListAgentsRequest{})
-	if err != nil {
-		return "", fmt.Errorf("client.SyntheticsAdmin.ListAgents: %w", err)
-	}
-
-	if getAllResp.GetAgents() != nil {
-		for _, agent := range getAllResp.GetAgents() {
-			if agent.GetType() == "private" {
-				return agent.GetId(), nil
-			}
-		}
-	}
-	return "", fmt.Errorf("no private agent found: %w", err)
-}
-
 func demonstrateSyntheticsTestsAPI() error {
 	ctx := context.Background()
 	client, err := NewClient()
@@ -165,7 +148,11 @@ func demonstrateSyntheticsTestsAPI() error {
 	fmt.Println("Number of invalid tests:", getAllResp.InvalidTestsCount)
 
 	fmt.Println("### Creating hostname synthetic test")
-	test := newHostnameTest()
+	test, err := newHostnameTest(ctx, client)
+	if err != nil {
+		return fmt.Errorf("new hostname test: %w", err)
+	}
+
 	test, err = client.Synthetics.CreateTest(ctx, test)
 	if err != nil {
 		return fmt.Errorf("client.SyntheticsAdmin.CreateTest: %w", err)
@@ -215,12 +202,22 @@ func demonstrateSyntheticsTestsAPI() error {
 	return nil
 }
 
-func newHostnameTest() *synthetics.Test {
+func newHostnameTest(ctx context.Context, client *kentikapi.Client) (*synthetics.Test, error) {
+	getAllResp, err := client.Synthetics.GetAllAgents(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("client.Synthetics.GetAllAgents: %w", err)
+	}
+
+	agentID, err := pickDualIPRustAgentID(getAllResp.Agents)
+	if err != nil {
+		return nil, fmt.Errorf("pick agent ID for hostname test: %w", err)
+	}
+
 	test := synthetics.NewHostnameTest(synthetics.HostnameTestRequiredFields{
 		BasePingTraceTestRequiredFields: synthetics.BasePingTraceTestRequiredFields{
 			BaseTestRequiredFields: synthetics.BaseTestRequiredFields{
 				Name:     "go-sdk-example-hostname-test",
-				AgentIDs: []string{"2592"},
+				AgentIDs: []string{agentID},
 			},
 			Ping: synthetics.PingSettingsRequiredFields{
 				Timeout:  10 * time.Second,
@@ -271,7 +268,7 @@ func newHostnameTest() *synthetics.Test {
 	test.Settings.Ping.Port = 65535
 	test.Settings.Traceroute.Port = 1
 
-	return test
+	return test, nil
 }
 
 func createMinimalTests() error {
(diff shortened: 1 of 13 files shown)


examples/synthetics_example_test.go
```diff
@@ -135,133 +147,400 @@ func pickPrivateAgentID(ctx context.Context) (string, error) {
...
+	fmt.Printf("Deleted synthetic test %q successfully\n", test.Name)
+	return nil
+}
+
+func newMinimalIPTest() *synthetics.Test {
+	return synthetics.NewIPTest(synthetics.IPTestRequiredFields{
+		BasePingTraceTestRequiredFields: synthetics.BasePingTraceTestRequiredFields{
+			BaseTestRequiredFields: synthetics.BaseTestRequiredFields{
+				Name:     "go-sdk-example-minimal-ip-test",
+				AgentIDs: []string{"2592"},
```
mmac-m3a: If this is always executed against a "fake" server, then hardcoding agent IDs is fine. Otherwise, agent IDs need to be selected by matching available agent by some criteria.

examples/synthetics_example_test.go
```diff
@@ -135,133 +147,400 @@ func pickPrivateAgentID(ctx context.Context) (string, error) {
...
+	return synthetics.NewHostnameTest(synthetics.HostnameTestRequiredFields{
+		BasePingTraceTestRequiredFields: synthetics.BasePingTraceTestRequiredFields{
+			BaseTestRequiredFields: synthetics.BaseTestRequiredFields{
+				Name:     "go-sdk-example-minimal-hostname-test",
+				AgentIDs: []string{"2592"},
+			},
+			Ping: synthetics.PingSettingsRequiredFields{
+				Timeout:  10 * time.Second,
+				Count:    10,
+				Protocol: synthetics.PingProtocolTCP,
```
mmac-m3a, line 413: Is the `port` set somewhere to a usable value?

examples/synthetics_example_test.go
```diff
@@ -135,133 +147,400 @@ func pickPrivateAgentID(ctx context.Context) (string, error) {
...
+			},
+			Traceroute: synthetics.TracerouteSettingsRequiredFields{
```
mmac-m3a, line 416: Same here.

examples/synthetics_example_test.go
```diff
@@ -135,133 +147,400 @@ func pickPrivateAgentID(ctx context.Context) (string, error) {
...
+		BaseTestRequiredFields: synthetics.BaseTestRequiredFields{
+			Name:     "go-sdk-example-minimal-page-load-test",
+			AgentIDs: []string{"2666"},
+		},
+		Definition: synthetics.TestDefinitionDNSRequiredFields{
+			Target:     "www.example.com",
+			Timeout:    time.Minute,
+			RecordType: synthetics.DNSRecordAAAA,
+			Servers:    []net.IP{net.ParseIP("192.0.2.213"), net.ParseIP("2001:db8:dead:beef:dead:beef:dead:beef")},
+			Port:       443,
```
mmac-m3a: Not very realistic port for DNS (which defaults to 53).

examples/synthetics_example_test.go
```diff
@@ -135,133 +147,400 @@ func pickPrivateAgentID(ctx context.Context) (string, error) {
...
+		BaseTestRequiredFields: synthetics.BaseTestRequiredFields{
+			Name:     "go-sdk-example-minimal-page-load-test",
+			AgentIDs: []string{"2666"},
+		},
+		Definition: synthetics.TestDefinitionDNSGridRequiredFields{
+			Target:     "www.example.com",
+			Timeout:    time.Minute,
+			RecordType: synthetics.DNSRecordAAAA,
+			Servers:    []net.IP{net.ParseIP("192.0.2.213"), net.ParseIP("2001:db8:dead:beef:dead:beef:dead:beef")},
+			Port:       443,
```
mmac-m3a: Same here.

kentikapi/internal/cloud/cloud.go
```diff
@@ -15,7 +15,7 @@ type API struct {
 	client cloudexportpb.CloudExportAdminServiceClient
 }
 
-// NewAPI creates new API.
+// NewAPI creates new cloud API.
```
mmac-m3a: More precisely "cloudexports" API.

kentikapi/internal/synthetics/test.go
```diff
@@ -0,0 +1,652 @@
...
+		Ping:                 pingSettingsFromPayload(ts.Ping),
+		Traceroute:           tracerouteSettingsFromPayload(ts.Trace),
+		Period:               time.Duration(ts.Period) * time.Second,
+		Family:               synthetics.IPFamily(ts.Family.String()),
+		NotificationChannels: ts.NotificationChannels,
+	}, nil
+}
+
+// nolint: gocyclo
+func testDefinitionFromPayload(t *syntheticspb.Test) (synthetics.TestDefinition, error) {
```
mmac-m3a, line 120: There is also `bgp_monitor` test type which should be silently ignored (it will be handled in separate BGP monitoring API).

mmac-m3a: The API may already return `transaction` type test, but it cannot yet provide actual configuration or results.

kentikapi/internal/synthetics/test.go
```diff
@@ -0,0 +1,652 @@
...
+	case synthetics.TestTypeDNS:
+		return dnsTestDefinitionFromPayload(t.Settings)
+	case synthetics.TestTypeDNSGrid:
+		return dnsGridTestDefinitionFromPayload(t.Settings)
+	default:
+		return nil, fmt.Errorf("invalid test type: %v", t.Type)
+	}
+}
+
+func ipTestDefinitionFromPayload(ts *syntheticspb.TestSettings) (synthetics.TestDefinition, error) {
```
mmac-m3a, line 147: Can go support general (introspection based) [construction of SDK objects based on protobuf data](https://github.com/kentik/community_sdk_python/blob/1bac5553fc04ba3d536602a372cd380c3ba32f87/kentik_api_library/kentik_api/synthetics/synth_tests/base.py#L56) like Python does?

kentikapi/internal/synthetics/test.go
```diff
@@ -0,0 +1,652 @@
...
+	}
+
+	return &syntheticspb.Test{
+		Id:            t.ID,
+		Name:          t.Name,
+		Type:          string(t.Type),
+		Status:        syntheticspb.TestStatus(syntheticspb.TestStatus_value[string(t.Status)]),
+		Settings:      ts,
+		Cdate:         nil, // read-only
+		Edate:         nil, // read-only
```
mmac-m3a: We need to support the ability to provide `edate` in `TestUpdate` requests.  If provided, the API returns error if the object to be modified has been modified in the DB after the `edate` timestamp. This allows to guard against concurrent modifications.

kentikapi/internal/synthetics/test.go
```diff
@@ -0,0 +1,652 @@
...
+		CreatedBy:     nil, // read-only
+		LastUpdatedBy: nil, // read-only
```
mmac-m3a, line 437: Tests now also support `labels`. Either add the support now, or create a ticket for later.

kentikapi/synthetics/test.go
```diff
@@ -0,0 +1,756 @@
...
+	// Settings contains test configuration attributes.
+	Settings TestSettings
+
+	// Read-only properties
+
+	// ID is unique test identification. It is read-only.
+	ID models.ID
+	// CreateDate is the creation timestamp. It is read-only.
+	CreateDate time.Time
+	// UpdateDate is the lst modification timestamp. It is read-only.
```
mmac-m3a: Typo: "lst" -> "last"

mmac-m3a (approved):
I'd like to get your feedback to my comments before merging the code. The code is good to go in as long as we've tickets for:
- support for labels for tests and agents
- support for passing `edate` in `TestUpdate` requests

kentikapi/internal/cloud/export.go
```diff
@@ -207,7 +211,7 @@ func cePayloadWithProperties(payload *cloudexportpb.CloudExport, ce *cloud.Expor
 	case ibmProvider:
 		payload.Properties = ibmPropertiesToPayload(ce)
 	default:
-		return nil, kentikerrors.New(kentikerrors.InvalidRequest, fmt.Sprintf("invalid cloud provider: %v", ce.Provider))
+		return nil, fmt.Errorf("invalid cloud provider: %v", ce.Provider)
```
mateuszmidor: As the list of supported providers may grow and cause error, maybe "unsupported cloud provider" woud better describe the case

kentikapi/internal/convert/convert.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"fmt"
+	"net"
+	"time"
+)
+
+func MillisecondsF32ToDuration(ms float32) time.Duration {
+	// scale to nanoseconds before conversion to duration to minimise conversion loss
+	const nanosPerMicro = 1e6
```
mateuszmidor: nanosPerMilli, probably

mateuszmidor (commented):
Looks good, minor comments left

examples/synthetics_example_test.go
danielfurman: Done: 2e37189
Examples are always executed against production server. Added note about that in readme and modified code to randomly select agents based on criteria.

examples/synthetics_example_test.go
danielfurman: No, it was not. I added it to required fields and filled with value for TCP pings. For ICMP pings this field can still be omitted, because 0 is zero value for uint32.
89dcddc

examples/synthetics_example_test.go
danielfurman: I wrongly assumed that server applies default here.
I added default value 33434 for UDP and TCP trace, which hopefully makes sense.
f1827f5

examples/synthetics_example_test.go
danielfurman: Yeah, I didn't bother with realism too much. Changed to 53.
eba3175

examples/synthetics_example_test.go
danielfurman: Done: https://github.com/kentik/community_sdk_golang/commit/eba3175dbd258b7b6da6678f4ae4d95f5c430de4

kentikapi/internal/cloud/cloud.go
danielfurman: Fixed: e741aa1

kentikapi/internal/synthetics/test.go
danielfurman: Added code for bgp_monitor tests: 4e80fa9

danielfurman: Added code for transaction tests: 3d7e8d2

mmac-m3a:
`godot` days: `Error: Comment should end in a period (godot)`... Like anybody was [waiting for him](https://en.wikipedia.org/wiki/Waiting_for_Godot) to days that :-).

examples/synthetics_example_test.go
```diff
@@ -55,9 +70,10 @@ func demonstrateSyntheticsAgentAPI() error {
 	fmt.Println("Number of agents:", len(getAllResp.Agents))
 	fmt.Println("Number of invalid agents:", getAllResp.InvalidAgentsCount)
```
mmac-m3a: That should be printed in large blinking red, if the value is non-zero :). The number is actually of any use for Kentik engineers. A non-zero value indicates some bug in the Kentik system. Let's print it only if it is non-zero and add text suggesting to contact Kentik support. Something like: `Kentik API returned <N> invalid agents. Please, contact Kentik support.`.

examples/synthetics_example_test.go
```diff
@@ -55,9 +70,10 @@ func demonstrateSyntheticsAgentAPI() error {
...
 
-	agentID, err := pickPrivateAgentID(ctx)
+	// Pick a private agent, so it is possible to modify it
+	agentID, err := pickPrivateAgentID(getAllResp.Agents)
```
mmac-m3a, line 80: Why do you insist on using a private agent here? There are definitely synthetics customers that have no private agents and still may want to learn about the SDK by running/tweaking examples.

Using a public agent IMHO makes more sense. There is (almost) guaranteed to be at least one available and it allows Kentik Trial customers to easily test the Go SDK.

examples/synthetics_example_test.go
```diff
@@ -143,125 +138,519 @@ func demonstrateSyntheticsTestsAPI() error {
...
-		return fmt.Errorf("client.SyntheticsAdmin.ListTests: %w", err)
+		return fmt.Errorf("client.Synthetics.GetAllTests: %w", err)
 	}
 
 	fmt.Println("Got all tests:", getAllResp)
-	fmt.Println("Number of tests:", len(getAllResp.GetTests()))
-	fmt.Println("Number of invalid tests:", getAllResp.InvalidCount)
-	fmt.Println()
+	fmt.Println("Number of tests:", len(getAllResp.Tests))
+	fmt.Println("Number of invalid tests:", getAllResp.InvalidTestsCount)
```
mmac-m3a: Like for agents, this number is of any use only to Kentik engineers.

examples/synthetics_example_test.go
```diff
@@ -143,125 +138,519 @@ func demonstrateSyntheticsTestsAPI() error {
...
+	if err != nil {
+		return err
+	}
+
+	getAllResp, err := client.Synthetics.GetAllAgents(ctx)
+	if err != nil {
+		return fmt.Errorf("client.Synthetics.GetAllAgents: %w", err)
+	}
+
+	dualIPRustAgentID, err := pickDualIPRustAgentID(getAllResp.Agents)
```
mmac-m3a: Considering that the goal of the example is to demonstrate the Go SDK (as opposed to test the Kentik API), let's use only IPv4 agents and tests. The reason for this is that the API is little bit behind the development of the Kentik synthetics system, and it currently cannot reliably represent dual stacked agents (it has no place to show the "other" IP address. Which one (v4 or v6) of addresses is shown in the `ip` (and also `local_ip`) property is undefined. The system internally (since about 2 months ago) does the right thing, but it is not visible in the API. The visible behavior may confuse more detailed oriented customers :).

We can add examples of choosing dual stack agents and picking protocol family in the test once the API catches up and we add support for new fields in the SDK.

examples/synthetics_example_test.go
```diff
@@ -143,125 +138,519 @@ func demonstrateSyntheticsTestsAPI() error {
...
+			Servers:    []net.IP{net.ParseIP("192.0.2.213"), net.ParseIP("2001:db8:dead:beef:dead:beef:dead:beef")},
+			Port:       53,
+		},
+	})
+}
+
+func pickPrivateAgentID(agents []synthetics.Agent) (models.ID, error) {
+	var matchedIDs []models.ID
+	for _, a := range agents {
+		if a.Type == synthetics.AgentTypePrivate {
```
mmac-m3a, line 577: Expecially for private agents, you need to also test that the `agent.status` is OK. If you want to be increase probability that the agent is actually operational, you can add test on the `last_authed` timestamp being less than 25 hours in the past (agents re-authorize every 24 hours).
Note that picking a non-operational agent causes test results to be unavailable, otherwise test creation (unfortunately) succeeds.

mmac-m3a (commented):
Getting ever better.

examples/synthetics_example_test.go
mmac-m3a: Note that soon (PR is pending review) node.js (`ksynth-agent`) agents will be allowed to run any synthetic test. Rust (`ksynth`) agents will never be able to run `page-load` and `transaction` tests.

kentikapi/internal/synthetics/test.go
danielfurman: Yes, I think it is possible, using the `reflect` package. There are several caveats, however:
1. I renamed some fields to be more user-friendly. That could be reverted or we could use struct tags on public models to specify PB field name.
2. Some PB fields are translated to user-friendly types, e.g. `string` to `ip.Addr`, `string` with number to `int`, `string` with duration to `time.Duration`. That could be reverted or we could use struct tags on public models to specify custom translation function.
3. I merged `activation.time_unit` and `activation.time_window` PB fields to single `Activation.TimeWindow` field of type time.Duration. That could be reverted or we could find some other mechanism to keep that state.
4. Possibly other aspects
5. There is a library for copying fields from one struct to other, but it might not be flexible enough: https://github.com/jinzhu/copier/  Its custom converters look messy: https://github.com/jinzhu/copier/pull/115

It would be quite hard to write and read. It might be more maintainable in the long run, but I am skeptical. I definitely don't want to do that in this PR. Do you want me to create a task for that?

kentikapi/internal/synthetics/test.go
danielfurman: Done: faa5fc9

kentikapi/synthetics/test.go
danielfurman: Done https://github.com/kentik/community_sdk_golang/commit/faa5fc9cb5bf42f91e09e96df25f8e05d5727bf2

kentikapi/internal/synthetics/test.go
danielfurman: Task created: https://clkentik.atlassian.net/browse/KNTK-409
It is not possible to do it now without manual modifications of generated proto structs, which we want to avoid.

kentikapi/internal/cloud/export.go
danielfurman: Done: d5528a9

kentikapi/internal/convert/convert.go
danielfurman: Nice catch; fixed

examples/synthetics_example_test.go
danielfurman: Done: 061b51c
Skipped the blinking red part ;)

examples/synthetics_example_test.go
danielfurman: For public agents, update and deactivate methods fail with `client.Synthetics.Deactivate: code: NotFound, message: rpc error: code = NotFound desc = could not find record (name:CustomError type:object) (errxid caxg9bnwhgf46mtema0g)`

From my point of view, the more methods are shown in the example, the better - both for users and maintainers. Especially as I am treating those examples as acceptance/integration tests.

I added a comment with instructions how to run the example without private agent: 2d46810 Is it a sufficient compromise?

examples/synthetics_example_test.go
danielfurman: Done: e9566f4

examples/synthetics_example_test.go
danielfurman: And fixed again: e9566f4

examples/synthetics_example_test.go
danielfurman: Modified example to use IPv4 agents: 695420c
Let us know when changes to agents arrive: we'll update docstrings and simplify `TestDemonstrateSyntheticsTestsAPI_CreateMinimalTests`

examples/synthetics_example_test.go
danielfurman: Ok, but this example (`demonstrateSyntheticsAgentsAPI`) only shows available methods of Agents API. It's purpose is not finding working agents, display test results, etc.

The same for examples for synth tests (`TestDemonstrateSyntheticsTestsAPI`): they only do some API manipulation and immediately clean up after.

danielfurman:
@mateuszmidor @mmac-m3a All comments addressed

kentikapi/internal/synthetics/test.go
mmac-m3a: Thanks for the explanation. We do not need to do it if the benefits are comparatively small.

examples/synthetics_example_test.go
mmac-m3a: Yes. I've missed that you are demonstrating agent activation and deactivation. Private agent is definitely needed for that.

examples/synthetics_example_test.go
mmac-m3a: OK.

mmac-m3a (approved):

PR #104, opened 2022-06-23
Improve synthetics and cloud code

danielfurman:
This commit includes improvements made while working on Synthetics Tests
interface:
- For existing cloud/synthetics code, move errors conversion up to API
  layer (from payload layer) to simplify the code and prevent omitting
  the conversion. Alongside that some error conversion bugs in cloud
  exports code were fixed.
- Remove Alibaba provider from cloud.Provider enum as it is not
  supported
- Generate String() method for Code type to fix its printing
- Describe stringer tool installation and code generation in readme.md
- Modify Start() function of fake servers to prevent flakiness
- Move spySyntheticsServer to separate file, because it is also going to
  be used in synth_test_integration_test.go
- Cosmetic improvements made while reading the code

Issue: KNTK-399


kentikapi/internal/cloud/export.go
```diff
@@ -304,66 +308,62 @@ func validateCEProvider(ce *cloud.Export) error {
...
 		return validateIBMProvider(ce)
 	default:
-		return kentikerrors.New(
-			kentikerrors.InvalidRequest,
-			fmt.Sprintf("cloud provider '%s' is not supported", ce.Provider))
+		return fmt.Errorf("cloud provider '%s' is not supported", ce.Provider)
 	}
 }
 
 func validateAWSProvider(ce *cloud.Export) error {
```
mmac-m3a, line 315: Could this repetitive code be possibly generalized or meta-programmed (i.e. generated from some template)?

mmac-m3a (approved):
LGTM. Nice cleanup. Please, consider my in-line comment for future improvements.

kentikapi/internal/cloud/export.go
danielfurman: Yes, it is possible.
1. We could write a function using reflections that would check required fields based on struct tags. We have something similar to that for checking nil pointers in API v5 request/response payloads: 
https://github.com/kentik/community_sdk_golang/blob/d23c4376f09d48d30357550569d0298f029081e7/kentikapi/internal/validation/validate.go#L18
It, however does not check for zero-values (only nil pointers).
2. Use a library for validation that also uses struct tags: https://github.com/go-playground/validator / https://github.com/asaskevich/govalidator
3. Use a library for validation that do not use struct tags: https://github.com/go-ozzo/ozzo-validation/

danielfurman: I added a note about that to synthetics validation task: https://clkentik.atlassian.net/browse/KNTK-402

PR #105, opened 2022-06-30
Provide consistent interface for synthetic tests results API

danielfurman:
- Move test results examples to one file
- Provide consistent interface for synthetic tests results API


mateuszmidor (approved):

kentikapi/internal/synthetics/test_results.go
```diff
@@ -0,0 +1,287 @@
...
+		EndTime:   timestamppb.New(req.EndTime),
+		AgentIds:  req.AgentIDs,
+		Targets:   convert.IPsToStrings(req.Targets),
+	}
+}
+
+type getResultsForTestsResponse syntheticspb.GetResultsForTestsResponse
+
+func (r *getResultsForTestsResponse) ToModel() ([]synthetics.TestResults, error) {
+	if r == nil {
```
mmac-m3a, line 31: This is actually a semi-normal state. There may not be any results available for synthetic tests for variety of reasons (not data collected yet, mis-configuration, non-functioning agents). We should return empty list instead of an error.

kentikapi/internal/synthetics/test_results.go
```diff
@@ -0,0 +1,287 @@
...
+			return nil, fmt.Errorf("agent results with index %v: %w", i, err)
+		}
+		ars = append(ars, ar)
+	}
+
+	return ars, nil
+}
+
+func agentResultsFromPayload(arPayload *syntheticspb.AgentResults) (synthetics.AgentResults, error) {
+	if arPayload == nil {
```
mmac-m3a, line 81: Similarly here. Agents may not have produced any data for a particular timestamp.

kentikapi/internal/synthetics/test_results.go
```diff
@@ -0,0 +1,287 @@
...
+
+func httpResponseDataFromPayload(d *syntheticspb.HTTPResponseData) (synthetics.HTTPResponseData, error) {
+	if d == nil {
+		return synthetics.HTTPResponseData{}, fmt.Errorf("HTTP response data is nil")
+	}
+
+	return synthetics.HTTPResponseData{
+		Status: d.Status,
+		Size:   d.Size,
+		Data:   d.Data,
```
mmac-m3a: The `data` attribute (if non-empty) actually contains JSON payload, which we should parse. Example:
```
  - id: '56114'
    health: critical
    tasks:
    - type: http
      target: http://e1-rp-nyc1.edgeuno.net [149.112.2.35]
      latency: 'current: 282.35ms avg: 281.62ms stddev: 0.723ms (healthy)'
      response:
        status: 404
        size: 153
        data:
          duration: 282346.5
          request_start: 142343
          https_validity: 0
          https_expiry_timestamp: 0
          errors: []
          connect_end: 142343
          domain_lookup_end: 2376.5
```

examples/synthetics_results_example_test.go
```diff
@@ -0,0 +1,422 @@
...
+		return fmt.Errorf("print ping packet loss matrix: %w", err)
+	}
+
+	return nil
+}
+
+func getLastTenTestResults(ctx context.Context, c *kentikapi.Client, testID string) ([]synthetics.TestResults, error) {
+	trs, err := c.Synthetics.GetResultsForTests(ctx, synthetics.GetResultsForTestsRequest{
+		TestIDs:   []models.ID{testID},
+		StartTime: time.Now().Add(-time.Hour * 240), // last 10 days
```
mmac-m3a: 10 days of results is 240 samples out of which you take 10. 1 day should be enough. A better solution is to get results for the last 12 test periods (which under normal circumstances guarantees 10 samples).

examples/synthetics_results_example_test.go
```diff
@@ -0,0 +1,422 @@
...
+// metricsMatrix holds mesh test results for ping task.
+type metricsMatrix struct {
+	// agents hold agents data in the same order as cells.
+	agents []synthetics.Agent
+	// cells hold all "fromAgent" -> "toAgent" connection metrics.
+	cells map[string]map[string]*synthetics.PingResults
+}
+
+func newMetricsMatrix(trs []synthetics.TestResults, allAgents []synthetics.Agent) (metricsMatrix, error) {
+	agents, err := prepareAgents(trs[0], allAgents)
```
mmac-m3a: Note that different time samples may have different number of agents reporting (e.g. due to configuration changes, agents being down or slowness in data collection). It is probably better to get the list of agents from the configuration of the test and ignore data from any other agents, leave blanks if results are not available.

mmac-m3a (changes requested):
Looks good, but we should not return error if no test results are available.

kentikapi/internal/synthetics/test_results.go
danielfurman: Changed: 6f3630033fddcd2de2fa3e849733bc67044b10ac

Note that this is a corner case when response object is not initialized. In current version of generated gRPC bindings it never happens, because response is initialized upfront: https://github.com/kentik/api-schema-public/blob/896729e59945404959e930f63f7a8965c109537d/gen/go/kentik/synthetics/v202202/synthetics_grpc.pb.go#L35

But I changed it anyway to be on the safe side.
Note that I return nil slices, because there is no need to initialize them - they work correctly with len(), append() and so on.

kentikapi/internal/synthetics/test_results.go
danielfurman: Changed: bd5dbf2d3609685b2b6d51932cc66a4f08922fe9

I didn't manage to hit that case by running examples / adding integration test but changed the code to ignore nil `*syntheticspb.AgentResults` without throwing error.

kentikapi/internal/synthetics/test_results.go
danielfurman: Done: 8ca507275e7c9dc21cbf57b69ff80d7169b58b2e

I modified the code to un-marshall data and convert map keys to snake case. Is it good enough?

examples/synthetics_results_example_test.go
danielfurman: Done: 92d1019
Changed to 12 hours. Should be good enough without complicating the example

examples/synthetics_results_example_test.go
danielfurman: I noticed that the example stopped working because of it. I'll fix it tomorrow.

danielfurman: Done: c7762d1
Unfortunately it doesn't work when test contains agents with dual IP - test results contain IP v4 and test configuration contain IP v6. So printed table has no data for connections to agents with IP v6.
It works well with IPv4 agents.

danielfurman:
@mmac-m3a All comments addressed

danielfurman:
I pushed [2b8e4e7](https://github.com/kentik/community_sdk_golang/pull/105/commits/2b8e4e767255debb14d7798decbc4b0ca6e82dd5) to fix failing example - some accepted timeout ranges in the API changed.

mmac-m3a (approved):

